PR #332, opened 2014-07-15
fix preloading of app with hidden pty with lots of output

ccutrer:
if we never read from the PTY, and the app spews a lot of messages
on startup, it might hang because the buffer is full


tmertens:
@jonleighton any chance we can get an official gem release with this fix soon?

This issue was causing all of our rake test commands to hang when run via spring (1.2.0).  The same rake tasks run fine on the curren spring master (ae175de3dd6dbae6d9aef59dbbfeed5a5eec90df).

jonleighton:
> any chance we can get an official gem release with this fix soon?

Done

PR #354, opened 2014-09-16
Do not clear GEM_HOME if there is any git-sourced gem in Gemfile.

### LINKED ISSUE #352: bundler options github and git causes warning
In one of my projects I got this warning from bundler every time I run the application:
```
git://github.com/grrrisu/Simulator.git (at master) is not yet checked out. Run `bundle install` first.
```
But the gem is installed correctly under $GEM_HOME/bundler/gems. Also **bundle list** shows the gem with the correct ref.

Every gem, that I require with with the **github** or **git** option in my Gemfile causes this warning. Digging down the problem, I discovered that unsetting the **ENV["GEM_HOME"]** in [bin/spring](https://github.com/rails/spring/blob/master/lib/spring/client/binstub.rb#L41) caused it. Now that no GEM_HOME is set, bundler tries to look up the gem in my_application/bundler/gems directory, which in my case does not exist and causes the warning. As soon as I remove the line from bin/spring, bundler works again as expected.

knu:
Fixes #352.

### DIFF
--- a/lib/spring/client/binstub.rb
+++ b/lib/spring/client/binstub.rb
@@ -36,12 +36,14 @@ unless defined?(Spring)
   require "rubygems"
   require "bundler"
 
-  if match = Bundler.default_lockfile.read.match(/^GEM$.*?^    (?:  )*spring \((.*?)\)$.*?^$/m)
+  case lockfile = Bundler.default_lockfile.read
+  when /^GEM$.*?^    spring \((.*?)\)$.*?^$/m
+    version = $1
     ENV["GEM_PATH"] = ([Bundler.bundle_path.to_s] + Gem.path).join(File::PATH_SEPARATOR)
-    ENV["GEM_HOME"] = ""
+    ENV["GEM_HOME"] = "" unless /^GIT$/ === lockfile
     Gem.paths = ENV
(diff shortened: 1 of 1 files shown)


monfresh:
I found a reproducible scenario, which I documented in issue #352. In my case, it seems to be caused by the `systemu` gem, which gets installed via the required dependencies of a gem in my Gemfile. @knu, do you happen to have the `systemu` gem in your Gemfile.lock?

knu:
@monfresh Thanks for tracking the problem down!  Looks like `systemu` is listed in [our Gemfile.lock](https://github.com/cantino/huginn/blob/master/Gemfile.lock); the gem is pulled in as dependency via `macaddr` via `uuid`.
We already [patched bin/spring locally](https://github.com/cantino/huginn/commit/53e2c8b4ede09498f8ea692ee0249bc930da6262#diff-6a60e1da6a5acd186690f18038f704b5) awhile ago and forgot about this PR, but it would be great if it helped.

jonleighton:
Hi guys, the setting of GEM_HOME has changed as of 1.3.0. It's now `nil` rather than `""`. Please give it a go.

Issue #368, opened 2014-12-03
binstubs sallowing unrealted LoadError


chancancode:
Today I ran into a problem that I had a LoadError somewhere in my Rails app. When I ran `bin/rails server`, this is what happened:

```
% bin/rails s
=> Booting WEBrick
=> Rails 4.2.0.rc1 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
=> Ctrl-C to shutdown server
Exiting
bin/rails:6: warning: already initialized constant APP_PATH
/private/tmp/loaderror/bin/rails:6: warning: previous definition of APP_PATH was here
Usage: rails COMMAND [ARGS]

The most common rails commands are:
 generate    Generate new code (short-cut alias: "g")
 console     Start the Rails console (short-cut alias: "c")
 server      Start the Rails server (short-cut alias: "s")
 dbconsole   Start a console for the database specified in config/database.yml
             (short-cut alias: "db")
 new         Create a new Rails application. "rails new my_app" creates a
             new application called MyApp in "./my_app"

In addition to those, there are:
 destroy      Undo code generated with "generate" (short-cut alias: "d")
 plugin new   Generates skeleton for developing a Rails plugin
 runner       Run a piece of code in the application environment (short-cut alias: "r")

All commands can be run with -h (or --help) for more information.
%
```

As you can see, it attempted to start the server, then silently failed, then followed up with some unrelated warnings, and finally a help message from the raw `rails` command complaining that I didn't pass any arguments.

If you would like to play along, I [pushed a sample application on github](https://github.com/chancancode/loaderror/commit/9927ef3b5b8ca74ae24cc3a23b2cf4b9e2579d37).

The reason is that we have a line in the `rails` binstub that [swallowed all `LoadError`](https://github.com/rails/spring/blob/master/lib/spring/client/binstub.rb#L17). Since this error didn't originate from loading spring like the binstub assumed, it is unsafe to continue execution in this case because the Rails environment was already loaded, ARGV poped, ...

Not sure what's the best fix. Perhaps we can rescue the `LoadError` within the `spring` binstub and raise a `SpringNotInstalled` error, then rescue only that in the `rails` binstub?

I think there are a few issues on the tracker that was caused by this, but it's hard to tell for sure...

sonalkr132:
fixed by: #444 and can be closed.

PR #370, opened 2014-12-10
Avoid a blanket rescue for LoadError

chancancode:
First stab for fixing #368. If it looks reasonable (the constant is defined at the right spot, the rescue is wrapping the right thing, etc) then I can clean it up :smile:

### DIFF
--- a/lib/spring/client/binstub.rb
+++ b/lib/spring/client/binstub.rb
@@ -14,7 +14,7 @@ module Spring
       LOADER = <<CODE
 begin
   load File.expand_path("../spring", __FILE__)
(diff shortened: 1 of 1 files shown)


jonleighton:
Hey @chancancode, do you fancy looking at adding a test for this?

chancancode:
Oops! I'll get back to this today!

chancancode:
Sorry, I forgot to comment. I tried to add a test last week but I couldn't get it to fail. If someone else wants to take that, feel free to grab it. Otherwise I'll get back to this in my next block of free time!

PR #372, opened 2014-12-11
Read a .spring.rb from the current directory

carpodaster:
Spring already supports reading from `~/.spring.rb` and `config/spring.rb`. This PR adds a local `.spring.rb` in the current directory to the list of config files.

Rationale: `~/.spring.rb` is too global and `config/spring.rb` leaks into a Rails engine's host app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Next version
 
+* Read a `.spring.rb` from the local directory in addition to a global `~/.spring.rb`
 * Automatically restart spring after new commands are added. This means
   that you can add spring-commands-rspec to your Gemfile and then
   immediately start using it, without having to run `spring stop`.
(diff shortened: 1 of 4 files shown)


jonleighton:
Sorry for the slow reply.

> In a Rails engine, the Gem's top level directories like app and config are magic (afaik). I.e. you would put localization files that your gem provides into config/locales/ and they are auto-loaded with the engine. I haven't checked if that is the case for any *.rb file in there. It is my understanding that they are and having the spring.rb there doesn't feel right for that reason. I would rather have it outside any of these auto-loaded paths.

I don't think it's any different to a normal Rails app? There are things that get automatically loaded from `config/` in a normal Rails app, but it's not *any* file. (Subdirectories like `config/initializers/` and `config/locales/` are 'special' but Spring doesn't use them.) It's very common for gems to put config files in `config/`.

On this basis I don't want to add the extra complexity of a third config file location. If you have new information I'm happy to reconsider but at the moment I don't see the rationale.

PR #382, opened 2015-02-09
help texts suggests install spring-commands-* gems

matthucke:
"spring rspec" and other commands supported by "spring-commands-*" plugins are unknown by default, and the path forward is not obvious to a beginner.

This change simply adds a line to the help output when one of these commands is attempted:

Hint: you probably want to add the 'spring-commands-foo' gem

for any foo currently in the spring-commands-* list in the README.  For all other unknown commands, the output is unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.3.1
+
+* Improved help message - suggests installing appropriate plugin when a
+  known but uninstalled command is attempted.
+
 ## 1.3.0
 
 * Automatically restart spring after new commands are added. This means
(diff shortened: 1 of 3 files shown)


jonleighton:
I'm not keen on the idea of specifically listing "known plugins". Perhaps you can just modify the help message to say "Additional commands can be installed via plugins: https://github.com/rails/spring#additional-commands"

matthucke:
OK - should I rip out the bit where it recognizes plugin names from that hard-coded list?  And simply always print that message as part of the standard help text, regardless of what command name invoked it?

thanks

jonleighton:
Sounds good

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.3.1
```
jonleighton: This should say "Next release" since we don't know what the version number will be yet

matthucke:
I'll work on it at home tonight or tomorrow - thanks.

PR #385, opened 2015-02-14
Add note about FactoryGirl

andyw8:
Since FactoryGirl is such a common gem I think it's worth a mention here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ provide a speed up on platforms which don't support forking (Windows, JRuby).
 
 ## Interoperability with other gems
 
-* FactoryGirl 4.4.1 or newer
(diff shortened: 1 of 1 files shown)


stgeneral:
Seems there is no sense in such README section at all. Spring does not have any influence on most of the gems.

andyw8:
Perhaps I wasn't clear. The author of Spring [reported the problem in FactoryGirl](https://github.com/thoughtbot/factory_girl_rails/issues/120), and it was fixed in thoughtbot/factory_girl_rails@b44b6b106b5275b7c6e1329a1530329547ed64d8. If someone adds Spring to an app running an older version of FactoryGirl then it will likely cause confusion when the factories aren't reloaded.

jonleighton:
Thanks, but since the fixed version has been out for over a year I don't really want to add this to the readme. Perhaps it could be suitable for a wiki page or something.

PR #393, opened 2015-03-03
Plural version of the model detected, using singularized version.

xiangzhuyuan:


jonleighton:
Why change this?

xiangzhuyuan:
Since model name should keep in singularized version, if we use a plural format, rails generator will raise a warning to us.

xiangzhuyuan:
although it will be ok for rails generator, but why we not following the conventions.

PR #398, opened 2015-03-23
remove warnings "warning: shadowing outer local variable - name"

y-yagi:
This removes the following warning:
```
spring-1.3.3/lib/spring/client/binstub.rb:149: warning: shadowing outer local variable - name
```


y-yagi:
Thanks!

PR #400, opened 2015-03-31
Add rails test command

arthurnn:
review @rafaelfranca @dhh 

I tried testing this in a rails app I have using `gem 'spring', '~> 1.3.3', github: 'arthurnn/spring'` on my gemfile, but for some reason it was saying that the test command didnt exist.
I am not sure if it is my setup or something, what it looked like was that the spring server was not running the new code, so it didnt have the test command or something like that.

### DIFF
--- a/lib/spring/commands/rails.rb
+++ b/lib/spring/commands/rails.rb
@@ -83,9 +83,16 @@ module Spring
       end
     end
 
+    class RailsTest < Rails
+      def command_name
+        "test"
+      end
+    end
+
     Spring.register_command "rails_console",  RailsConsole.new
     Spring.register_command "rails_generate", RailsGenerate.new
     Spring.register_command "rails_destroy",  RailsDestroy.new
     Spring.register_command "rails_runner",   RailsRunner.new
+    Spring.register_command "rails_test",     RailsTest.new
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Not released
+
+* Add `rails test` command.
+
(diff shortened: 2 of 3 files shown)


arthurnn:
cc @senny

lib/spring/commands/rails.rb
```diff
@@ -83,9 +83,16 @@ def extract_environment(args)
...
+      def command_name
+        "test"
+      end
+    end
+
     Spring.register_command "rails_console",  RailsConsole.new
     Spring.register_command "rails_generate", RailsGenerate.new
     Spring.register_command "rails_destroy",  RailsDestroy.new
     Spring.register_command "rails_runner",   RailsRunner.new
+    Spring.register_command "rails_test",   RailsTest.new
```
rafaelfranca: I tihnk we should align it

dhh:
I think @eileencodes just had the same issue. Resolution was to nuke the spring process.

eileencodes:
Yea I haven't looked into why yet but I just ripped spring out "to be sure" and I had the test command. Spring stop wasn't working. I had to remove it from the Gemfile.

lib/spring/commands/rails.rb
arthurnn: right.. done!

arthurnn:
I was able to test this locally, and it works.
The trick is, you cannot use `gem 'spring', path: ''` or `gem 'spring' github: ''` . See https://github.com/rails/spring/blob/master/CONTRIBUTING.md#testing-with-your-app

dhh:
Mucho gracias!

PR #402, opened 2015-04-15
Lock the version of the spawned server

### LINKED ISSUE #295: Client-server version mismatch when multiple versions of spring installed for the current Ruby
I've noticed that when using multiple versions of Spring in the app's current Ruby, I get that error. This is a common occurrence because we have many apps on the same Ruby version, but possibly using dfferent versions of Spring. Even doing `bundle exec bin/rails c` or `bundle exec spring rails c` results in the same behavior.

```
$ bin/rails c
...
...
There is a version mismatch between the spring client and the server.
You should restart the server and make sure to use the same version.

CLIENT: 1.0.0, SERVER: 1.1.2
```

TimMoore:
This fixes a bug when the currently-executing version of Spring is not the most recent version available in the gem path. Previously, this would launch the most recent server version and cause an error when Bundler detects the mismatch.

Fixes #295.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 1.3.5
 
 * Fix `rails test` command to run in test environment #403 - @eileencodes
+* Ensure the spawned server is loaded from the same version of the Spring gem
+  as the client. Issue #295.
 
 ## 1.3.4
 
--- a/lib/spring/client/run.rb
+++ b/lib/spring/client/run.rb
@@ -68,8 +68,7 @@ module Spring
         pid = Process.spawn(
           gem_env,
           "ruby",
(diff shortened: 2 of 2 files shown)


TimMoore:
Any thoughts about writing a test for this? It's a little tricky to add a regression test, because the bug only manifests when you're running an older version of the gem than the most recent version installed.

TimMoore:
Is there any interest in this change from the maintainers? I'm happy to rebase the branch and fix the merge conflicts if so.

jonleighton:
Hi @TimMoore, sorry for the slowness. This change looks good to me. Please do rebase and I'll get it released.

TimMoore:
Thanks, @jonleighton. Done!

Issue #396, opened 2015-03-19
Spring hangs a lot


mockdeep:
We've been having a lot of problems with spring hanging on our systems. I started outputting to a log and here's what I see:

```
[2015-03-19 14:27:01 -0700] [21417] [application:test] initialized -> running
/home/fletch/.rvm/gems/ruby-2.1.5/gems/spring-1.3.3/lib/spring/application.rb:286:in `read': stream closed (IOError)
    from /home/fletch/.rvm/gems/ruby-2.1.5/gems/spring-1.3.3/lib/spring/application.rb:286:in `block (2 levels) in with_pty'
[2015-03-19 14:27:01 -0700] [12838] [application_manager:test] child 21417 shutdown
[2015-03-19 14:27:01 -0700] [21531] [application:test] preloading app
[2015-03-19 14:27:06 -0700] [21531] [application:test] initialized -> running
/home/fletch/.rvm/gems/ruby-2.1.5/gems/spring-1.3.3/lib/spring/application.rb:286:in `read': stream closed (IOError)
    from /home/fletch/.rvm/gems/ruby-2.1.5/gems/spring-1.3.3/lib/spring/application.rb:286:in `block (2 levels) in with_pty'
```

I haven't got a strong idea of what is causing it just yet, though it seems to come up if we change git branches and the like, so maybe some sort of file changes are causing it to just lock up. We have to run `spring stop` and start over.

ulfmagnetics:
+1 also seeing a lot of hangs on OS X Yosemite.

mockdeep:
I'm on Linux Mint, though other users at our company are seeing it on Mac as well. Something as simple as changing branches can cause it:

```
[2015-03-26 12:20:36 -0700] [10899] [application:development] running -> watcher_stale
[2015-03-26 12:20:36 -0700] [10899] [application:development] watcher_stale -> exiting
[2015-03-26 12:20:36 -0700] [10896] [application_manager:development] child 10899 shutdown
[2015-03-26 12:20:36 -0700] [11147] [application:development] preloading app
[2015-03-26 12:20:42 -0700] [11147] [application:development] initialized -> running
/home/fletch/.rvm/gems/ruby-2.1.5/gems/spring-1.3.3/lib/spring/application.rb:286:in `read': stream closed (IOError)
	from /home/fletch/.rvm/gems/ruby-2.1.5/gems/spring-1.3.3/lib/spring/application.rb:286:in `block (2 levels) in with_pty'
[...]

PR #404, opened 2015-05-11
Sets verbosity level to 'silence' for applications

jafrog:
I'm not sure if this is good as a default option, but I believe it would solve a lot of cases of [this issue](https://github.com/rails/spring/issues/396) (see my comment on the issue).


jonleighton:
Thanks for researching this issue, however this doesn't seem like an ideal fix to me, and I imagine that it might not fix all scenarios where that bug might occur. Therefore I'd prefer to hold out for a proper fix which prevents spring from hanging in the first place (rather than preventing the situation where spring hangs). Perhaps you could try to write a test case for this scenario as a first step?

PR #414, opened 2015-08-01
speed up tests by not copying bundled directory every time

grosser:
I noticed that the tests run super slow ... because we copy all the gems on every run.

`generator.copy_to(app.root)` takes about 2-3s on my machine,
with this patch it's kept in test/app/gems which makes the copy just take 0.04s

before: 4m33s
after: 2m51s

@rafaelfranca @amatsuda @jonleighton

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: ruby
+cache: bundler
+sudo: false
 rvm:
   - 1.9.3
   - 2.0.0
--- a/lib/spring/test/acceptance_test.rb
+++ b/lib/spring/test/acceptance_test.rb
@@ -194,9 +194,13 @@ module Spring
       end
 
       test "binstub when spring is uninstalled" do
-        app.run! "gem uninstall --ignore-dependencies spring"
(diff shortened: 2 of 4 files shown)


grosser:
ok, got that fixed, 1 test was uninstalling spring

grosser:
travis times old:
screen shot 2015-08-01 at 9 29 20 pm

new:
screen shot 2015-08-01 at 9 29 28 pm

grosser:
another commit to kick out a few gems we don't need (sdoc/web-console/byebug), which will decreases first time install time (new users + travis), travis was not impressed, but locally I can see install time drop + runtime of `time rails runner 'puts 1'` decrease by 0.4s

and another to use travis docker infrastructure -> more cpu -> faster builds to gets rid of this warning:
screen shot 2015-08-01 at 10 21 54 pm

also now caches the installed gems (not the apps folder, just the root projects Gemfile) to get rid of these 6s 
image

for the finally runtime of:

image

jonleighton:
Nice work. Happy to merge this when the commented-out tests pass too.

grosser:
ups ... that was leftover debugging, updated :D

PR #423, opened 2015-08-15
Fix STDOUT/ERR kept open after preload failure

### LINKED ISSUE #422: stdout won't be closed when error occured in preload
## Summary

When error is occurred while preloading rails app, with stdout piped to another command:

- Expected: exit and return to console
- Actual: piped command does not exit until `spring stop` is run

I found this issue while developing [auto completion plugin for text editor](https://github.com/ypresto/robe-atom).

## Gem versions

rails 4.2.3
spring 1.3.6

## Steps to reproduce

1. rails new test-rails-app
1. Insert syntax error in config/application.rb (e.g. remove `end` of last line)
1. bin/rails runner 'some script' | tail
1. SyntaxError stack trace is printed out, but tail does not exit.
1. bin/spring stop will exit tail command and return to console.

This issue does not happen when syntax error is placed in 'some script', not application.rb.

ypresto:
Fixes #422

Thanks!

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -192,7 +192,6 @@ module Spring
       }
 
       disconnect_database
-      reset_streams
 
       log "forked #{pid}"
       manager.puts pid
@@ -209,6 +208,11 @@ module Spring
 
       client.puts(1) if pid
       client.close
+    ensure
+      # Redirect STDOUT and STDERR to prevent from keeping the original FDs
+      # (i.e. to prevent `spring rake -T | grep db` from hanging forever),
+      # even when exception is raised before forking (i.e. preloading).
+      reset_streams
     end
 
     def terminate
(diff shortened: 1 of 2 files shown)


grosser:
Looks great, I was expecting worse hackery :D 
can you add a test-case with this scenario so this does not break in the future ?

jonleighton:
Yeah, a test would be good please. Also needs a changelog entry. Thanks.

ypresto:
It is very difficult to implement test case because it should use separated process and should be end-to-end test.

jonleighton:
@ypresto most of our tests are end-to-end tests so there is already a good structure for it. There's already a [very similar test](https://github.com/rails/spring/blob/master/lib/spring/test/acceptance_test.rb#L414-L416) to the one that is needed here.

ypresto:
Thanks, added test..!

PR #426, opened 2015-08-29
support pre-boot hooks

grosser:
@jonleighton @rafaelfranca 

a solution for https://github.com/grosser/parallel_tests/issues/412

also allows a few hacks I had in mind:
 - make stop pgrep and kill all matching processes (to fix random orphans I saw)
 - add `start` so you have to `spring start` in a separate tab and see the log output there

(obviously needs tests etc, but direction looks good ?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,9 @@ settings. Note that `~/.spring.rb` is loaded *before* bundler, but
 projects without having to be added to the project's Gemfile, require
 them in your `~/.spring.rb`.
 
+`config/spring_preboot.rb` is also loaded before bundler and before a 
+server process is started, it can be used to add new top-level commands.
+
 ### Application root
 
 Spring must know how to find your Rails application. If you have a
--- a/lib/spring/client.rb
+++ b/lib/spring/client.rb
@@ -36,3 +36,9 @@ module Spring
     end
   end
 end
+
+# allow users to add hooks that do not run in the server
+# or modify start/stop
+if File.exist?("config/spring_preboot.rb")
+  require "./config/spring_preboot.rb"
+end
--- a/lib/spring/test/acceptance_test.rb
+++ b/lib/spring/test/acceptance_test.rb
@@ -262,6 +262,13 @@ module Spring
         assert_success "bin/rails runner 'puts 2'", stdout: "!callback!\n2"
       end
 
+      test "can define preboot tasks" do
+        File.write("#{app.spring_config.sub('.rb', '_preboot.rb')}", <<-RUBY)
+          Spring::Client::COMMANDS["foo"] = lambda { |args| puts "bar -- \#{args.inspect}" }
+        RUBY
+        assert_success "bin/spring foo --baz", stdout: "bar -- [\"foo\", \"--baz\"]\n"
+      end
+
       test "missing config/application.rb" do
         app.application_config.delete
         assert_failure "bin/rake -T", stderr: "unable to find your config/application.rb"


rafaelfranca:
Yes, looks good to me

grosser:
added a test, looking good ?

grosser:
@webandtech  does this look like it would help you ?

rafaelfranca:
:shipit:

webandtech:
@grosser Not sure. I haven't circled back to this issue as we simply stayed put on parallel_tests 1.3.9.

grosser:
sry, mentioned you on the wrong project :D

grosser:
bumped to 1.4.0, can you release @rafaelfranca / @jonleighton ?

grosser:
FYI using this now:

 - keep a tab open to know 'spring is still running'
 - see log output to be able to quickly debug things
 - stops spring on ctrl+c -> easy cleanup / never forget spring is still running
 - block execution if start was not used

... have to turn that into a gem ... alternatively I could add it and trigger when `ENV['SPRING_START']` is defined ... @rafaelfranca / @jonleighton do you think this is useful and should be here or a gem ?

```Ruby
# listen to the spring log and stop on ctrl+c
class Spring::Client::Start < Spring::Client::Command
  LOG = 'log/spring.log'
  LOCK = 'tmp/spring.lock'

  def self.description
    "Tail #{LOG} and stop spring on Cltr+c"
  end

  def call
    if env.server_running?
      abort "Spring is already running"
    else
      puts "Spring ready to rumble!"
      exec "trap 'rm -rf #{LOCK} && spring stop && exit' 2; touch #{LOCK} && tail -f -n 0 #{LOG}"
    end
  end
end

Spring::Client::COMMANDS['start'] = Spring::Client::Start

# make everyone use spring start
class << Spring::Client
  prepend(Module.new do
    def run(args)
      if ["start", "help", "binstub", "stop", "status", "-v", "--version"].include?(args.first)
        super
      elsif File.exist?(Spring::Client::Start::LOCK)
        super
      else
        abort "use: spring start"
      end
    end
  end)
end

# always log
ENV['SPRING_LOG'] = Spring::Client::Start::LOG
```

rafaelfranca:
It is indeed useful. Lets start as a gem in the same way we did wit spring commands.

README.md
```diff
@@ -294,6 +294,9 @@ settings. Note that `~/.spring.rb` is loaded *before* bundler, but
 projects without having to be added to the project's Gemfile, require
 them in your `~/.spring.rb`.
 
+`config/spring_preboot.rb` is also loaded before bundler and before a 
+server process is started, it can be used to add new top-level commands.
```
jonleighton: Sorry for the late reply, but I think this might be better named `config/spring_client.rb` - that would make it clearer in my mind that there is a "client" and a "server" side to spring, and that this file allows modifying the client side (whereas the other config files operate on the server side).

jonleighton:
> do you think this is useful and should be here or a gem ?

It's worth saying that the idea of this command is kind of counter to Spring's original aims; i.e. I was trying to create something which would be a transparent as possible and "just work" without extra steps of starting/stopping etc. However it's clear that this goal was only partially achieved and some people do run into problems, or else others just prefer to have more visibility into what Spring is doing. So perhaps something like this would be useful, but I agree that it would be good to keep developing it outside spring for now and then consider including it by default when it's a bit more mature.

README.md
grosser: https://github.com/rails/spring/pull/431

PR #428, opened 2015-09-04
add bump to get rake bump:patch/minor/major tasks

grosser:
makes it easy (no need to search where the version is stored) and error-free to bump the version (changes + bundles + commits for you)

@rafaelfranca @jonleighton


rafaelfranca:
Oh. Didn't know about this gem. Nice one. :shipit:

PR #429, opened 2015-09-05
make help more accessible

grosser:
@jonleighton @rafaelfranca 

I found myself using `spring -h` twice already ... would be nice if it was consistent with other commands and did not try to start a server (and keep it running in the background) when I use `-h` / `--help` because I might be a new user and not even know that springs wants `spring help`

### DIFF
--- a/lib/spring/test/acceptance_test.rb
+++ b/lib/spring/test/acceptance_test.rb
@@ -84,6 +84,14 @@ module Spring
 
       test "help message when called without arguments" do
         assert_success "bin/spring", stdout: 'Usage: spring COMMAND [ARGS]'
+        assert app.spring_env.server_running?
+      end
+
+      test "shows help" do
+        assert_success "bin/spring help", stdout: 'Usage: spring COMMAND [ARGS]'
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Yeah agree :+1:

grosser:
tests added, looking good ?

lib/spring/test/acceptance_test.rb
```diff
@@ -84,6 +84,13 @@ def without_gem(name)
 
       test "help message when called without arguments" do
         assert_success "bin/spring", stdout: 'Usage: spring COMMAND [ARGS]'
+        assert app.spring_env.server_running?
+      end
+
+      test "shows help" do
+        assert_success "bin/spring help", stdout: 'Usage: spring COMMAND [ARGS]'
+        assert_success "bin/spring --help", stdout: 'Usage: spring COMMAND [ARGS]'
```
rafaelfranca: Missing a test of `-h`?

grosser: added!

rafaelfranca:
:shipit:

grosser:
can you release a 1.4.0 ?

PR #430, opened 2015-09-05
Watch files autoloaded by AS::Dependencies during initialization.

oggy:
Without this, modules that are loaded during initialization, and stored in a global reference (including modules mixed in to core or library classes) would not trigger a spring restart when modified.

A simple demo of the problem is below. Although `TestConstant` is reloaded correctly when modified, `TEST_VALUE` remains stale. Another (perhaps more common) example of this pattern is mixing a module into a core class or a class from a gem during initialization. In that case, it can lead to the puzzling "A copy of X has been removed from the module tree but is still active!" error.
## Demo

Apply to a new Rails app:

``` diff
diff --git a/config/application.rb b/config/application.rb
index 11b2ec6..a57aa19 100644
--- a/config/application.rb
+++ b/config/application.rb
@@ -22,5 +22,7 @@ module Springy

     # Do not swallow errors in after_commit/after_rollback callbacks.
     config.active_record.raise_in_transactional_callbacks = true
+
+    config.autoload_paths << "#{Rails.root}/lib"
   end
 end
diff --git a/config/initializers/test.rb b/config/initializers/test.rb
new file mode 100644
index 0000000..b34f716
--- /dev/null
+++ b/config/initializers/test.rb
@@ -0,0 +1,2 @@
+TEST_VALUE = TestConstant
+
diff --git a/lib/test_constant.rb b/lib/test_constant.rb
new file mode 100644
index 0000000..eb99609
--- /dev/null
+++ b/lib/test_constant.rb
@@ -0,0 +1 @@
+TestConstant = 2
```

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -103,8 +103,7 @@ module Spring
       watcher.add e.backtrace.map { |line| line.match(/^(.*)\:\d+\:in /)[1] }
       raise e unless initialized?
     ensure
-      watcher.add loaded_application_features
-      watcher.add loaded_application_dependencies
+      watcher.add loaded_application_files
       watcher.add Spring.gemfile, "#{Spring.gemfile}.lock"
 
       if defined?(Rails) && Rails.application
@@ -235,8 +234,7 @@ module Spring
     def setup(command)
       if command.setup
         # Loaded features may have changed.
-        watcher.add loaded_application_features
-        watcher.add loaded_application_dependencies
+        watcher.add loaded_application_files
       end
     end
 
@@ -244,6 +242,10 @@ module Spring
       Spring.after_fork_callbacks.each do |callback|
         callback.call
       end
+    end
+
+    def loaded_application_files
+      loaded_application_features + loaded_application_dependencies
     end
 
     def loaded_application_features
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -1,6 +1,7 @@
 require "spring/boot"
 require "set"
 require "pty"
+require "rbconfig"
 
 module Spring
   class Application
@@ -103,6 +104,7 @@ module Spring
       raise e unless initialized?
     ensure
       watcher.add loaded_application_features
+      watcher.add loaded_application_dependencies
       watcher.add Spring.gemfile, "#{Spring.gemfile}.lock"
 
       if defined?(Rails) && Rails.application
@@ -232,7 +234,9 @@ module Spring
     # load the helper file once and have it cached.
     def setup(command)
       if command.setup
-        watcher.add loaded_application_features # loaded features may have changed
+        # Loaded features may have changed.
+        watcher.add loaded_application_features
+        watcher.add loaded_application_dependencies
       end
     end
 
@@ -245,6 +249,18 @@ module Spring
     def loaded_application_features
       root = Spring.application_root_path.to_s
       $LOADED_FEATURES.select { |f| f.start_with?(root) }
+    end
+
+    def loaded_application_dependencies
+      root = Spring.application_root_path.to_s
+      loaded = ActiveSupport::Dependencies.loaded
+
+      # The loaded_paths do not include file suffixes. Find all matches.
+      dlexts = RbConfig::CONFIG.values_at('DLEXT', 'DLEXT2')
+      ext_glob = "{rb,#{dlexts.reject { |s| s.to_s.empty? }.join(',')}}"
+      loaded.select { |path| path.start_with?(root) }.flat_map do |path|
+        Dir["#{path}.#{ext_glob}"]
+      end
     end
 
     def disconnect_database
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -256,10 +256,11 @@ module Spring
       loaded = ActiveSupport::Dependencies.loaded
 
       # The loaded_paths do not include file suffixes. Find all matches.
-      dlexts = RbConfig::CONFIG.values_at('DLEXT', 'DLEXT2')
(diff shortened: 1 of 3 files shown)


grosser:
Sounds like funky problem, great to see you dug in deep to fix it :)
can you add a test too so this does not break in the futue ?

oggy:
Ah, missed the acceptance tests under `lib`. Added!

grosser:
hmm ci seems to be failing ... do the tests pass locally ?

lib/spring/application.rb
```diff
@@ -247,6 +251,18 @@ def loaded_application_features
       $LOADED_FEATURES.select { |f| f.start_with?(root) }
     end
 
+    def loaded_application_dependencies
+      root = Spring.application_root_path.to_s
+      loaded = ActiveSupport::Dependencies.loaded
+
+      # The loaded_paths do not include file suffixes. Find all matches.
+      dlexts = RbConfig::CONFIG.values_at('DLEXT', 'DLEXT2')
+      ext_glob = "{rb,#{dlexts.reject { |s| s.to_s.empty? }.join(',')}}"
```
grosser: if there are no dlext then this would be `{.rb,}` is that right ?

maybe `extensions = ['.rb'] + RbConfig::CONFIG.values_at('DLEXT', 'DLEXT2').reject(&:empty)`

lib/spring/test/acceptance_test.rb
```diff
@@ -104,15 +104,14 @@ def without_gem(name)
         end
       end
 
-      test "code changes in pre-referenced app files are picked up" do
```
grosser: was the old test bad ? <-> leave it and add a new one ? (smaller tests are usually easier to fix/maintain)

lib/spring/application.rb
oggy: Good catch.

lib/spring/test/acceptance_test.rb
oggy: The old test was kind of obsolete. The old test checked that the changed file will be reloaded. The new behavior is that spring will restart in this case. The old test would fail because the spring restart would not result in the expected speedup.

oggy:
Pushed a fix for the no-dlext case (not actually sure if that ever happens, but robustness is good).

The tests are passing for me locally (`bundle exec rake`). Travis may be having a slow day -- guessing if we were to bump that timeout on CI it'd pass.

lib/spring/application.rb
```diff
@@ -232,7 +234,9 @@ def exit_if_finished
     # load the helper file once and have it cached.
     def setup(command)
       if command.setup
-        watcher.add loaded_application_features # loaded features may have changed
```
grosser: hmm same as the above 2 lines, can we dry this up this into a method ?

grosser:
@jonleighton / @rafaelfranca  looking good to you too ?


Tried an empty build and it worked: https://travis-ci.org/rails/spring/jobs/78933263
I could not restart this PR's build ... can you ? (https://travis-ci.org/rails/spring/builds/78933260) ... hopefully it's green now ... but maybe there is something bad going on ...

lib/spring/test/acceptance_test.rb
rafaelfranca: That is not good. This means that if you are configuring a model in an initializer and you change the model the spring process will be restarted where before it would not.

rafaelfranca:
Restarted the build and it is broken (you should be able to restart too, I think you only need to request access on rails org in travis settings).

I'm not sure about this. I raised my concerns at https://github.com/rails/spring/pull/430#discussion_r39013076

oggy:
Thanks @rafaelfranca . I'm convinced there is something here causing the remaining CI failure. Will investigate.

Regarding your concern:

The confusing thing to me is that if you explicitly require a file in an initializer, then edits to the file do trigger a restart, but if you load it implicitly by referencing the constant, then it does not. Restarting in both cases seems like the safer alternative: worst case you restart when you don't need to; the other way you can get puzzling errors (as mentioned up top).

If you're configuring a model, and not updating any state outside that model, then I'm thinking it's better to not load it in an initializer at all, at least in dev/test. Better to have the module load its configuration when loaded, and keep app startup faster.

jonleighton:
Hi @oggy, thanks for the PR.

Personally I'm not in favour of this change, because I think it could lead to people wondering why they're "not getting the speed-up". I think the current situation is simpler to reason about:

* File in `autoload_paths` changes: constants are reloaded
* File outside `autoload_paths` changes: full application restart

This also mirrors the autoloading situation which people are used to when using development web servers etc. In your example, I don't think `TEST_VALUE` would change when using `rails server`, even if you changed `TestConstant`.

In my view it's just a behaviour that we have to explain and deal with. Our docs go into some detail about it:

https://github.com/rails/spring#class-reloading

Of course, if you can think of a way to make the error messages more useful, that would be great.

oggy:
@jonleighton

Thanks for the reply, and the documentation pointer.

I see where you're coming from, but I still find the current rules more confusing. Perhaps because I don't view spring as something like `rails server`, but rather as a thing that wraps `rails server` (and other commands) to give a transparent, faster speed-ups. Transparent being the key here. The first bullet under "Features" in the readme seems to echo this: "Totally automatic; no need to explicitly start and stop the background process" -- which is currently not quite true.

Documenting the edge case is better than nothing, but I'm thinking better would be to emit a warning when the case is detected. And if a warning is present, it seems preferable to warn about potential slowness and do the reload, rather than warning that you might need to manually restart if you see a funky error at some later point.

There are pros & cons -- let me know your thoughts and I can update the PR.

rafaelfranca:
Rails 6 changed the way it load code and this PR is now stale. I agree with job that we should just explain this behavior. Thank you

PR #437, opened 2015-09-19
Prefer single-quoted strings

mariusbutuc:
RuboCop's [Style/StringLiterals](http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Style/StringLiterals) recommends to:

> Prefer single-quoted strings when you don't need string interpolation or special symbols

I understand that the camps are divided around nitpicks like this one, but this minimal and inexpensive change—in both the docs and the generated code—can reduce number of notices a new user would get.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ code into relevant existing executables. The snippet looks like this:
 
 ``` ruby
 begin
-  load File.expand_path("../spring", __FILE__)
+  load File.expand_path('../spring', __FILE__)
 rescue LoadError
 end
 ```
@@ -295,7 +295,7 @@ settings. Note that `~/.spring.rb` is loaded *before* bundler, but
 projects without having to be added to the project's Gemfile, require
 them in your `~/.spring.rb`.
 
-`config/spring_client.rb` is also loaded before bundler and before a 
+`config/spring_client.rb` is also loaded before bundler and before a
 server process is started, it can be used to add new top-level commands.
 
 ### Application root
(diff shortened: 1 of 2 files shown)


mariusbutuc:
Or, I could [quote @fxn](https://github.com/rails/rails/commit/83bbc2352c73beba655d9c1d58a3244782209bc5) :smile: 

> prefer single quotes, by the beard of the prophet!

grosser:
I don't really care either way ... but want to throw in this: https://viget.com/extend/just-use-double-quoted-ruby-strings

afaik there is no significant performance benefit ... avoiding a few rubocop warnings for whoever pastes code is nice ... but I've also seen projects use double-quoted style ... so meh :D

jonleighton:
Personally, I prefer double quotes, because there's no performance benefit to single quotes, and I hate having to convert the quoting style just to be able to use interpolation.

That said, this PR is really about what style we use in the *generated* code. If the rest of the code generated by Rails uses single quotes (does it?), we should also probably use single quotes for consistency I suppose.

fxn:
Yes, Rails has that convention (and the reasoning is not performance, I don't know who started thay myth).

rafaelfranca:
:+1: for the change in the generated code, but not for spring code. @mariusbutuc could you update your pull request to change only the generated code?

rafaelfranca:
Looks like everything in this PR is in the generated code. Thanks!

PR #438, opened 2015-10-13
Include CHANGELOG in gem

dbackeus:
Make it easier for developers to track changes after `bundle update` etc.


rafaelfranca:
In fact I'm planning to kill the CHANGELOG file in favor of GitHub releases. CHANGELOGs are a huge release pain because people expect us to keep different stable branches CHANGELOGs in sync. This is not needed when using GitHub releases.

Thank you for the pull request.

grosser:
:heart: oh yeah :D

PR #441, opened 2015-11-10
Add WINCH to FORWARDED_SIGNALS

CGamesPlay:
This is necessary, in conjunction with a change in ~/.irbrc, to enable terminal window resizing while using spring.

    Signal.trap('SIGWINCH', proc { Readline.set_screen_size(*`stty size`.split.map(&:to_i)); Readline.refresh_line })


grosser:
thx, looks good :)

grosser:
v1.4.1

PR #442, opened 2015-11-10
Fix the CHANGELOG entry from rails/spring#441

lucasmazza:
The entry was added under the existing `1.4.0` release.


kaspth:
Thanks 😁

PR #444, opened 2015-11-15
Don't supress non-spring load errors in binstub

### LINKED ISSUE #259: warning: already initialized constant APP_PATH
I'm getting a `warning: already initialized constant APP_PATH` error, coming from `bin/rails`. I'm using the latest Spring and Rails 4.

felixbuenemann:
Without this change, the spring binstub supresses unrelated load errors, eg. due to errors in initializers or loaded gems. This change causes the binstub to only ignore a missing spring loader or gem.

This will get rid of the misleading `warning: already initialized constant APP_PATH` and resolves #259.

### DIFF
--- a/lib/spring/client/binstub.rb
+++ b/lib/spring/client/binstub.rb
@@ -13,8 +13,10 @@ module Spring
       # should cause the "unsprung" version of the command to run.
       LOADER = <<CODE
 begin
-  load File.expand_path('../spring', __FILE__)
-rescue LoadError
+  spring_bin_path = File.expand_path('../spring', __FILE__)
+  load spring_bin_path
+rescue LoadError => e
+  raise unless e.message.end_with? spring_bin_path, 'spring/binstub'
 end
 CODE
 
--- a/lib/spring/test/acceptance_test.rb
+++ b/lib/spring/test/acceptance_test.rb
@@ -208,10 +208,19 @@ module Spring
         assert_success "bin/rake -T", stdout: "rake db:migrate"
       end
 
-      test "binstub when spring is uninstalled" do
+      test "binstub when spring gem is missing" do
         without_gem "spring-#{Spring::VERSION}" do
           File.write(app.gemfile, app.gemfile.read.gsub(/gem 'spring.*/, ""))
           assert_success "bin/rake -T", stdout: "rake db:migrate"
+        end
+      end
+
+      test "binstub when spring binary is missing" do
+        begin
+          File.rename(app.path("bin/spring"), app.path("bin/spring.bak"))
+          assert_success "bin/rake -T", stdout: "rake db:migrate"
+        ensure
+          File.rename(app.path("bin/spring.bak"), app.path("bin/spring"))
         end
       end
 


felixbuenemann:
The acceptance tests currently only cover the case of a missing spring gem, but not of a missing spring binary. If a missing spring binary should blow up with a load error, I could omit the check for spring_loader.

felixbuenemann:
Hmm, seems like Ruby 1.9.3 doesn't have `LoadError#path`, I could check the exception message instead.

felixbuenemann:
OK, updated to use `LoadError#message` instead of `#path`.

grosser:
can you add an acceptance test that fails without this patch ? (like loading a random gem and asserting it blows up nicely)

lib/spring/client/binstub.rb
```diff
@@ -13,8 +13,9 @@ class Binstub < Command
       # should cause the "unsprung" version of the command to run.
       LOADER = <<CODE
 begin
-  load File.expand_path('../spring', __FILE__)
-rescue LoadError
+  load spring_loader = File.expand_path('../spring', __FILE__)
```
grosser: doing the assignment on an extra line would improve readability a bit

grosser:
I think that's a great change/pattern, too many libraries do this rescue all ...

chancancode:
I guess this supersedes https://github.com/rails/spring/pull/370 :grin: :clap:

Can you find out if there are any relevant comment from Jon and add "Closes #370" to your commit message?

Thanks for picking this up!

lib/spring/client/binstub.rb
felixbuenemann: Yes, but it would also break the existing regex to check for the old/new style binstub.

felixbuenemann:
@chancancode Oh, I didn't see your PR because it wasn't linked to the issue mentioned above.

I think we could add a test to check for the case were the spring binary is missing.

@grosser What do you think should be the intended behavior, when the `spring` binary is missing? (Example: There's a springified bin/rails, but bin/spring has been deleted.) Should a LoadError be raised, or should it be ignored (current but maybe unintended behavior).

grosser:
is there any downside to ignoring it ? wrong version will be used / slower ?

lib/spring/client/binstub.rb
grosser: how is this breaking a regex ?
```Ruby
spring_loader = File.expand_path('../spring', __FILE__)
load spring_loader
```

felixbuenemann:
Well, it should only happen, if a user deleted `bin/spring` without restoring the original binstubs. There are no performance problems and given that there doesn't seem to be a command to "unspringify" the binstubs we should probably handle that case.
My initial thought for not ignoring it, was to alert the user that their spring setup is broken. While the second case (missing gem) is expected, eg. because spring is only loaded in the development environment.

grosser:
So sounds like you think it's best to ignore it ?

felixbuenemann:
Yes, I'll add a test that makes the expected behavior explicit.

lib/spring/client/binstub.rb
felixbuenemann: Good catch, both `load File.expand_path('../spring', __FILE__)` and `load spring_loader` match the regex `/load .*spring/` in `Spring::Binstub::Item#add`, so it's OK.

felixbuenemann:
Adressed feedback from @grosser and @chancancode and added a new test that checks the behavior in absence of a `bin/spring` binary.

grosser:
:tada:

grosser:
1.4.2 is out !

felixbuenemann:
Great, thanks Michael!

PR #445, opened 2015-11-17
Fix upgrade/remove with new binstub variations

felixbuenemann:
This is a followup to #444 which adds support for upgrading to the new loader code as well as supporting `bin/spring binstub --remove` with both single and double quoted variants of the previous loader code, which was not handled in #437.

I've also added an acceptance test for `bin/spring binstub --remove --all` which tests for the removal of `bin/spring` and added a mention of the `--remove` option to the spring client binstub help output.

### DIFF
--- a/lib/spring/client/binstub.rb
+++ b/lib/spring/client/binstub.rb
@@ -119,7 +119,7 @@ CODE
       attr_reader :bindir, :items
 
       def self.description
-        "Generate spring based binstubs. Use --all to generate a binstub for all known commands."
+        "Generate spring based binstubs. Use --all to generate a binstub for all known commands. Use --remove to revert."
       end
 
       def self.rails_command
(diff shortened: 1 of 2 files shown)


felixbuenemann:
@rafaelfranca This should probably be assigned to @grosser, because he handled #444.

lib/spring/client/binstub.rb
```diff
@@ -48,6 +48,12 @@ class Binstub < Command
 
       OLD_BINSTUB = %{if !Process.respond_to?(:fork) || Gem::Specification.find_all_by_name("spring").empty?}
 
+      BINSTUB_VARIATIONS = Regexp.union [
+        %{begin\n  load File.expand_path("../spring", __FILE__)\nrescue LoadError\nend\n},
```
grosser: these 2 lines look pretty similar ?

grosser: ah the quotes ...

felixbuenemann: Double vs. single quotes.

grosser:
great tests! :)

grosser:
1.4.3 since it's kind of a fix for the previous ... even though it includes a new feature ... but should be safe ...

felixbuenemann:
I agree, this is a bug fix release. Thanks for taking the time to review.

PR #446, opened 2015-11-17
Mention spring-commands-rubocop gem

p0deje:
Mentions https://github.com/p0deje/spring-commands-rubocop as a gem which adds `rubocop` command.


grosser:
thx, the more the merrier ;)

PR #448, opened 2015-11-18
Fix uncaught Gem::LoadError in binstub loader

### LINKED ISSUE #447: spring is not part of the bundle for production environment
It seems that new binstub format (introduced with #444) is causing the issue in the production envrionment when the bundle is installed without development and test:

````
/app/vendor/bundle/ruby/2.2.0/gems/bundler-1.9.7/lib/bundler/rubygems_integration.rb:263:in `block in replace_gem': spring is not part of the bundle. Add it to Gemfile. (Gem::LoadError)
	from /app/bin/spring:12:in `<top (required)>'
	from /app/bin/rake:4:in `load'
	from /app/bin/rake:4:in `<main>'
````

Full setup:
rails 2.4.5 application
ruby 2.2.3
spring 1.4.3
heroku cedar-14 stack
bundle is installed using `bundle install --without development:test --path vendor/bundle --binstubs vendor/bundle/bin -j4 --deployment`

The issue occurs when running `heroku run rake db:migrate`

### PR DESCRIPTION
**Pull request by felixbuenemann**
The previous binstub code added in #444 did not catch Gem::LoadErrors which can be generated by bundler or rubygems trying to load the spring gem.

This fixes #447.

### DIFF
--- a/lib/spring/client/binstub.rb
+++ b/lib/spring/client/binstub.rb
@@ -13,10 +13,9 @@ module Spring
       # should cause the "unsprung" version of the command to run.
       LOADER = <<CODE
 begin
-  spring_bin_path = File.expand_path('../spring', __FILE__)
-  load spring_bin_path
+  load File.expand_path('../spring', __FILE__)
 rescue LoadError => e
-  raise unless e.message.end_with? spring_bin_path, 'spring/binstub'
+  raise unless e.message.include?('spring')
 end
 CODE
 
@@ -51,6 +50,7 @@ CODE
       BINSTUB_VARIATIONS = Regexp.union [
         %{begin\n  load File.expand_path("../spring", __FILE__)\nrescue LoadError\nend\n},
         %{begin\n  load File.expand_path('../spring', __FILE__)\nrescue LoadError\nend\n},
+        %{begin\n  spring_bin_path = File.expand_path('../spring', __FILE__)\n  load spring_bin_path\nrescue LoadError => e\n  raise unless e.message.end_with? spring_bin_path, 'spring/binstub'\nend\n},
         LOADER
       ]
 
--- a/lib/spring/test/acceptance_test.rb
+++ b/lib/spring/test/acceptance_test.rb
@@ -275,6 +275,13 @@ module Spring
       end
 
       test "binstub upgrade with new binstub variations" do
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by felixbuenemann** [comment]
?r @grosser

File: lib/spring/client/binstub.rb [hunk 1]
```diff
@@ -13,10 +13,9 @@ class Binstub < Command
       # should cause the "unsprung" version of the command to run.
       LOADER = <<CODE
 begin
-  spring_bin_path = File.expand_path('../spring', __FILE__)
-  load spring_bin_path
+  load File.expand_path('../spring', __FILE__)
 rescue LoadError => e
-  raise unless e.message.end_with? spring_bin_path, 'spring/binstub'
+  raise unless e.message =~ /spring/
```
grosser: `.include?("spring")` ?

lib/spring/client/binstub.rb
```diff
@@ -51,6 +50,7 @@ class Binstub < Command
       BINSTUB_VARIATIONS = Regexp.union [
         %{begin\n  load File.expand_path("../spring", __FILE__)\nrescue LoadError\nend\n},
         %{begin\n  load File.expand_path('../spring', __FILE__)\nrescue LoadError\nend\n},
+        %{begin\n  spring_bin_path = File.expand_path('../spring', __FILE__)\n  load spring_bin_path\nrescue LoadError => e\n  raise unless e.message.end_with? spring_bin_path, 'spring/binstub'\nend\n},
```
grosser: this is getting annoying :D

lib/spring/client/binstub.rb
felixbuenemann: Should suffice, I'll update.

lib/spring/client/binstub.rb
felixbuenemann: Well it's much easier now that the variations are listed and used by both upgrade and remove.

grosser: yeah, that was a great refactor :D

felixbuenemann:
Updated to use `include?` instead of regex.

felixbuenemann:
Btw. I've manually tested against heroku with the new binstub code.

grosser:
1.4.4

Issue #275, opened 2014-03-20
Dependency on Bundler not stated


strzibny:
Hi,

spring seems to have a dependency on Bundler, but this dependency is nowhere no be found. I understand that bundler is pulled in with Rails (and that spring is used together with Rails), but one can run Rails easily without Bundler.

Can we add it as a dependency? (I can make a PR if you want to)

Also can somebody explain to me why Bundler is really needed? Or is there a way to use spring without Bundler?

Thank you.

PR #453, opened 2015-12-10
Add bundler Dependency to gemspec

sonalkr132:
`bundle exec` was added to acceptance test cause otherwise it was giving Gem::LoadError for conflicting bundler versions.

An alternate solution would have been to add it [here](https://github.com/rails/spring/blob/master/lib/spring/test/application.rb#L94-L103):

```Ruby
 Bundler.with_clean_env do
   Process.spawn(
     env,
     command.to_s,   # may be change this line to "bundle exec #{command}"
     out:   stdout.last,
     err:   stderr.last,
     in:    :close,
     chdir: root.to_s,
   )
 end
```
I felt it was non-intuitive. Additionally, above code block doesn't get called only from `Spring::Test::AcceptanceTest`, so it would need changes on many different files.

**EDIT:**
Resolves: #275

### DIFF
--- a/lib/spring/test/application.rb
+++ b/lib/spring/test/application.rb
@@ -94,7 +94,7 @@ module Spring
         Bundler.with_clean_env do
           Process.spawn(
             env,
-            command.to_s,
+            "bundle exec #{command}",
             out:   stdout.last,
             err:   stderr.last,
             in:    :close,
@@ -200,7 +200,7 @@ module Spring
       end
 
       def bundle
-        run! "(gem list bundler | grep bundler) || gem install bundler", timeout: nil, retry: 2
+        run! "gem list bundler | grep bundler || gem install bundler", timeout: nil, retry: 2
         run! "bundle check || bundle update --retry=2", timeout: nil
       end
 
--- a/spring.gemspec
+++ b/spring.gemspec
@@ -13,6 +13,7 @@ Gem::Specification.new do |gem|
   gem.files         = Dir["LICENSE.txt", "README.md", "lib/**/*", "bin/*"]
   gem.executables   = gem.files.grep(%r{^bin/}).map{ |f| File.basename(f) }
 
+  gem.add_dependency 'bundler', '>= 1.0'
   gem.add_development_dependency 'activesupport', '~> 4.2.0'
   gem.add_development_dependency 'rake'
   gem.add_development_dependency 'bump'
--- a/lib/spring/test/application_generator.rb
+++ b/lib/spring/test/application_generator.rb
@@ -85,9 +85,9 @@ module Spring
       def install_spring
         return if @installed
 
-        build_and_install_gems
-
         application.bundle
+
+        build_and_install_gems
 
         FileUtils.rm_rf application.path("bin")
 


lib/spring/test/acceptance_test.rb
```diff
@@ -83,14 +83,14 @@ def without_gem(name)
       end
 
       test "help message when called without arguments" do
-        assert_success "bin/spring", stdout: 'Usage: spring COMMAND [ARGS]'
+        assert_success "bundle exec bin/spring", stdout: 'Usage: spring COMMAND [ARGS]'
```
grosser: we always advertise to not use bundle exec ... so with_clean_env would be better :(

sonalkr132: I think error on travis (check [this](https://travis-ci.org/sonalkr132/spring/jobs/95848491)) can be on user side as well (?). Looking at the test won't help him a lot cause it won't be apparent that command passing in test is not actually:
```Ruby
bin/spring
```
but 
```Ruby
bundle exec bin/spring
```

Also, **some** of the commands being passed to `with_clean_env` are:
```Ruby
# gives syntax error near unexpected token `('
# can be solved by just removing '('
(gem list bundler | grep bundler) || gem install bundler
# not sure from where does this gets passed some jugglery will be needed    
bundle check || bundle update --retry=2 

## some gem install commands
## more commands from other files calling the same app.run 
```

sonalkr132: @grosser Do you really want me to change it to `with_clean_env`?
Actually, it needs just 2 lines of change.

grosser: I'm confused now :D 
Do you think this is better or with_clean_env is better ?
I think either would work, I was just sad to see bundle exec being added all over the place + that it uses a pattern we try to discourage ...

sonalkr132: `with_clean_env` feels like we are secretly adding `bundle exec` to every command.
However, `with_clean_env` is lot cleaner and gets the job done. Also, issue of conflicting bundler version seems to be only specific for travis runs. Locally, I have bundler 1.10.6 and I don't get any error.

sonalkr132:
@grosser I changed it to `with_clean_env` version. Can you please have a look?
Order of `application.bundle` and `build_and_install_gems` is changed cause to bundle and install gems we will need bundler first. Bundler is install by `application.bundle`.

grosser:
looks good to me ... but could break all kinds of stuff ... @jonleighton want to have a look ?

sonalkr132:
> but could break all kinds of stuff 

Are referring to change of order of `application.bundle` and `build_and_install_gems`? Those things are only used for tests and has nothing to do with working of gem.

grosser:
ahhh you are right! :D 
so you don't need a new release after this ,right ?

sonalkr132:
> so you don't need a new release after this ,right?

TBH, I am not sure. It does add `bundler` as hard dependency.

sonalkr132:
Shit! I just noticed it. I did reset hard and forgot to add the most important line -_-

grosser:
:D

sonalkr132:
Fixed it!
> so you don't need a new release after this ,right?

I think it will need changelog that now spring has bundler has hard dependency.

grosser:
:+1: 

@jonleighton / @rafaelfranca looks good ?

lib/spring/test/application.rb
```diff
@@ -200,7 +200,7 @@ def run!(command, options = {})
       end
 
       def bundle
-        run! "(gem list bundler | grep bundler) || gem install bundler", timeout: nil, retry: 2
+        run! "gem list bundler | grep bundler || gem install bundler", timeout: nil, retry: 2
```
jonleighton: Why change this? It makes it less readable IMO

spring.gemspec
```diff
@@ -13,6 +13,7 @@ Gem::Specification.new do |gem|
   gem.files         = Dir["LICENSE.txt", "README.md", "lib/**/*", "bin/*"]
   gem.executables   = gem.files.grep(%r{^bin/}).map{ |f| File.basename(f) }
 
+  gem.add_dependency 'bundler', '~> 1.3'
```
jonleighton: I think we can make this more permissive. Perhaps `~> 1.0` is fine.

jonleighton: In fact, `>= 1.0`

lib/spring/test/application.rb
```diff
@@ -94,7 +94,7 @@ def run(command, opts = {})
         Bundler.with_clean_env do
           Process.spawn(
             env,
-            command.to_s,
+            "bundle exec #{command}",
```
jonleighton: Why is this necessary? Which tests fail without it?

We are already setting `GEM_PATH` and `GEM_HOME` in the env of the spawned process, so I don't understand why it would be an issue.

lib/spring/test/application.rb
sonalkr132: `bundle exec (gem list bundler | grep bundler) || gem install bundler` gives following error:

> bash: syntax error near unexpected token `('

lib/spring/test/application.rb
sonalkr132: > We are already setting GEM_PATH and GEM_HOME in the env of the spawned process, so I don't understand why it would be an issue.

Yes, it shouldn't be an issues and it isn't, at least locally. However, tests on travis fail with conflicting bundler version. For *some* reason [travis installs bundler 1.7.6](https://travis-ci.org/sonalkr132/spring/jobs/96552775#L101) at first and later it ends up conflicting with generated rails app bundler. Same thing happened in last two attempts of this fix ( #293, #373).

sonalkr132:
Can someone please suggest tests so that #456 doesn't happen again? I can't think of anything reasonable.

sonalkr132:
closed in favor of #458

PR #455, opened 2015-12-12
print spring client pid on command run

### LINKED ISSUE #407: Print notice when client process uses Spring
I just burned an afternoon because something started Spring and I had no idea it was running.  It would have saved me hours if `bundle exec rails c` had printed "Using spring process 12345" when it started up so I'd known it was active.

I hear people complain about debugging mysterious problems due to out-of-date code preloaders all the time so it would seem this is a pretty common issue.  Would it be possible to add a notice?

sonalkr132:
Resolves #407

### DIFF
--- a/bin/spring
+++ b/bin/spring
@@ -46,4 +46,6 @@ end
 lib = File.expand_path("../../lib", __FILE__)
 $LOAD_PATH.unshift lib unless $LOAD_PATH.include?(lib) # enable local development
 require 'spring/client'
+# if the user knows that spring is called, do not show running spring
+Spring.quiet = true if $0.include? 'spring'
 Spring::Client.run(ARGV)
--- a/lib/spring/client/run.rb
+++ b/lib/spring/client/run.rb
@@ -1,3 +1,4 @@
+require "spring/commands"
 require "rbconfig"
 require "socket"
 require "bundler"
@@ -137,6 +138,8 @@ ERROR
 
         if pid && !pid.empty?
           log "got pid: #{pid}"
+
+          puts "Running via Spring preloader in process #{pid}" unless Spring.quiet
 
           forward_signals(pid.to_i)
           status = application.read.to_i
--- a/lib/spring/configuration.rb
+++ b/lib/spring/configuration.rb
@@ -2,7 +2,7 @@ require "spring/errors"
 
 module Spring
   class << self
-    attr_accessor :application_root
+    attr_accessor :application_root, :quiet
 
     def gemfile
       ENV['BUNDLE_GEMFILE'] || "Gemfile"
@@ -49,4 +49,6 @@ module Spring
       end
     end
   end
+
+  self.quiet = false
 end
--- a/lib/spring/test/acceptance_test.rb
+++ b/lib/spring/test/acceptance_test.rb
@@ -94,6 +94,19 @@ module Spring
         refute app.spring_env.server_running?
       end
 
+      test "tells the user that spring is being used when used automatically via binstubs" do
+        assert_success "bin/rails runner ''", stdout: "Running via Spring preloader in process"
+        assert_success app.spring_test_command, stdout: "Running via Spring preloader in process"
+      end
+
+      test "does not tell the user that spring is being used when used automatically via binstubs but quiet is enabled" do
+        File.write("#{app.user_home}/.spring.rb", "Spring.quiet = true")
+        assert_success "bin/rails runner ''"
+        artifacts = app.run("bin/rails runner ''")
+        assert !artifacts[:stdout].include?("Running via Spring preloader in process"),
+          "expected stdout to not include 'Running via Spring preloader in process'.\n\n#{app.debug(artifacts)}"
+      end
+
       test "test changes are picked up" do
         assert_speedup do
           assert_success app.spring_test_command, stdout: "0 failures"
--- a/lib/spring/test/acceptance_test.rb
+++ b/lib/spring/test/acceptance_test.rb
@@ -447,7 +447,7 @@ module Spring
             system(#{app.env.inspect}, "bundle install")
           end
           output = `\#{Rails.root.join('bin/rails')} runner 'require "devise"; puts "done";'`
-          exit output == "done\n"
+          exit output.include? "done\n"
         RUBY
 
         assert_success [%(bin/rails runner 'load Rails.root.join("script.rb")'), timeout: 60]
--- a/lib/spring/test/acceptance_test.rb
+++ b/lib/spring/test/acceptance_test.rb
@@ -47,6 +47,14 @@ module Spring
         assert_output artifacts, expected_output if expected_output
       end
 
+      def refute_output_includes(command, not_expected)
+        artifacts = app.run(*Array(command))
+        not_expected.each do |stream, output|
+          assert !artifacts[stream].include?(output),
+                 "expected #{stream} to not include '#{output}'.\n\n#{app.debug(artifacts)}"
+        end
+      end
+
       def assert_speedup(ratio = DEFAULT_SPEEDUP)
         if ENV['CI']
           yield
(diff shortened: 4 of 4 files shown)


grosser:
how about: 
 - when started via `spring rails xxx` don't print
 - when started via `rails xxx` print by default ?

sonalkr132:
@grosser I don't know if I got this right, but for this we need to differentiate between two types of commands (namely `spring rails/rake xxx` and `rails/rake xxx`).
[This](https://github.com/rails/spring/blob/master/bin/spring#L49) line is probably the first line that gets called:
```Ruby
Spring::Client.run(ARGV)
```
ARGV is same for both cases (`spring rails c` and `rails c`). Probably cause `bin/rails` acts a stand-in and just delegates commands to `bin/spring` leaving no difference between two commands.
Can you think of a way to differentiate between two commands?

grosser:
$0 for bin/rails is ./bin/rails and for spring is rails_runner

sonalkr132:
done!

bin/spring
```diff
@@ -46,4 +46,5 @@ end
 lib = File.expand_path("../../lib", __FILE__)
 $LOAD_PATH.unshift lib unless $LOAD_PATH.include?(lib) # enable local development
 require 'spring/client'
+Spring.verbose_output = false if $0.include? 'spring' # don't output running spring status if command has `spring` in it
```
grosser: how about: `Spring.verbose_output = $0.include?('spring')`

lib/spring/test/acceptance_test.rb
```diff
@@ -47,6 +47,14 @@ def assert_failure(command, expected_output = nil)
         assert_output artifacts, expected_output if expected_output
       end
 
+      def assert_not_include(command, not_expected)
```
grosser: that should be `refute_include` ?
but that's already defined by minitest ...

grosser:
:+1: @jonleighton looks good ?

bin/spring
grosser: the comment repeats the code, how about:

`if the user knows that spring is called, do not show that spring is called`

lib/spring/test/acceptance_test.rb
sonalkr132: `assert_not_include` is [defined in test::unit](http://www.rubydoc.info/github/test-unit/test-unit/Test%2FUnit%2FAssertions%3Aassert_not_include).

`assert_output` [defined in minitest](http://ruby-doc.org/stdlib-1.9.3/libdoc/minitest/unit/rdoc/MiniTest/Assertions.html#method-i-assert_output) is also overloaded here. Should I change `assert_not_include` to `refute_include` anyway?

bin/spring
sonalkr132: > how about: Spring.verbose_output = $0.include?('spring')

I think you meant `Spring.verbose_output = !$0.include?('spring')`
It is not going to work it will set `Spring.verbose_output` to `true` when command includes`spring`, even when user has `Spring.verbose_output = false` defined in `config/spring.rb` (or `~/.spring.rb`).

grosser: ah yeah, makes sense

lib/spring/test/acceptance_test.rb
grosser: let's try to not override them :D
maybe something different like refute_output_includes or so ?

lib/spring/configuration.rb
```diff
@@ -49,4 +49,6 @@ def find_project_root(current_dir)
       end
     end
   end
+
+  self.verbose_output = true
```
jonleighton: Calling the option "verbose" makes me think that it produces more than the normal amount of output. However the default is "verbose" mode.

I'd call this `Spring.quiet` instead, and have it default to false. Then you're opting in to quietness, rather than the opting out of verbosity. (I also think the `_output` bit is superfluous.)

lib/spring/client/run.rb
```diff
@@ -1,7 +1,9 @@
+require "spring/commands"
 require "rbconfig"
 require "socket"
 require "bundler"
 
+
```
jonleighton: Unnecessary line added here

lib/spring/client/run.rb
```diff
@@ -138,6 +140,8 @@ def run_command(client, application)
         if pid && !pid.empty?
           log "got pid: #{pid}"
 
+          puts "Running Spring [client] #{pid}" if Spring.verbose_output
```
jonleighton: I'd rephrase this:

```
Running via Spring preloader in process 1234
```

* I'm not sure the `[client]` bit has any meaning to somebody who doesn't know about the details of Spring's implementation
* The "Spring preloader" bit can be put into a search engine to figure out what Spring is (for users who have installed Rails out-the-box and who don't already know)

jonleighton:
Made some comments but generally :+1:

sonalkr132:
@jonleighton @grosser Thanks for getting me through it :stuck_out_tongue_closed_eyes: 

I have made the changes you requested.

lib/spring/test/acceptance_test.rb
```diff
@@ -94,6 +102,22 @@ def without_gem(name)
         refute app.spring_env.server_running?
       end
 
+      test "running spring status for rails/rake commands" do
```
grosser: these descriptions could make it easier for other users to understand what is being tested:

```
test "tells the user that spring is being used when used automatically via binstubs"
test "does not tell the user that spring is being used when the user used spring"
test "does not tell the user that spring is being when used automatically via binstubs but quiet is set to true"
```

grosser:
thx, I'll merge when travis is green :)

lib/spring/test/acceptance_test.rb
sonalkr132: Thanks.

sonalkr132:
hey, a quick question: Is `watch_interval` config serving any purpose? It is mentioned in `lib/spring/watcher.rb` but never really used any where else in repo. Check [this](https://github.com/rails/spring/search?utf8=%E2%9C%93&q=watch_interval)

grosser:
looks unused ... can you make a cleanup PR to remove it ?

grosser:
here we got!

grosser:
have fun with 1.6.0  :)

sonalkr132:
yay :tada:

jonleighton:
> hey, a quick question: Is watch_interval config serving any purpose? It is mentioned in lib/spring/watcher.rb but never really used any where else in repo. Check this

I think I just put it there to avoid having magic numbers in the code. I don't think we should hardcode something like this.

sonalkr132:
> Is watch_interval config serving any purpose? It is mentioned in lib/spring/watcher.rb but never really used any where else in repo

> looks unused ... can you make a cleanup PR to remove it?

I was wrong that it is not serving any purpose. (check [this](https://travis-ci.org/sonalkr132/spring/jobs/96793847) travis run). 
@jonleighton is it defined in some library spring is using? Which class behavior does it affect?

lib/spring/client/run.rb
```diff
@@ -1,3 +1,4 @@
+require "spring/commands"
```
hsume2: @sonalkr132 hey, quick question. I was wondering why you added this line? I'm getting an exception which seems to go away when I remove this line (which I've logged [here](https://github.com/rails/spring/issues/457)).

sonalkr132: I had to require `spring/command` cause it [loads the user defined config files](https://github.com/rails/spring/blob/master/lib/spring/commands.rb#L28-L29) (`config/spring.rb` or `~/.spring.rb`).

If user has defined `Spring.quite` on one of those files then we need different behavior [here](https://github.com/rails/spring/pull/455/files#diff-c1a62c31f3c02a71c2d99bf2faee9457R142):

```
puts "Running via Spring preloader in process #{pid}" unless Spring.quiet
```

hsume2: Got it, thanks!

jonleighton: @sonalkr132 @grosser I missed this, but we shouldn't be requiring `spring/commands` in the client. apart from this issue that @hsume2 reported, loading bundler in the client slows down the execution of spring commands, which is exactly what we want to avoid. The client needs to run as quickly as possible.

For client-side configuration, there is the `config/spring_client.rb` file, so in theory the user could set `Spring.quiet` in there (although I don't think this is really something we need to document...)

After fixing this, it would be great if we could somehow prevent a similar bug from being reintroduced in the future (perhaps we can somehow write a test which asserts that bundler is not present in the client? I don't know)

sonalkr132: So,`Spring.quite` option can only be set in `spring_client.rb`?

jonleighton: With the current implementation, I think the answer should be yes. However I think that actually we could write this "Running via Spring preloader" message in `Application#serve`, by which point `spring/commands` would be loaded. So I think that may be a better option.

PR #459, opened 2015-12-21
update backtrace string format regex

keepcosmos:
Every backtrace strings do not include `:in`
http://ruby-doc.org/core-1.9.3/Exception.html#method-i-backtrace

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -99,7 +99,7 @@ module Spring
       @preloaded = :success
     rescue Exception => e
       @preloaded = :failure
-      watcher.add e.backtrace.map { |line| line.match(/^(.*)\:\d+\:in /)[1] }
+      watcher.add e.backtrace.map { |line| line[/^(.*)\:\d+/, 1] }
       raise e unless initialized?
     ensure
       watcher.add loaded_application_features


lib/spring/application.rb
```diff
@@ -99,7 +99,7 @@ def preload
       @preloaded = :success
     rescue Exception => e
       @preloaded = :failure
-      watcher.add e.backtrace.map { |line| line.match(/^(.*)\:\d+\:in /)[1] }
+      watcher.add e.backtrace.map { |line| line.match(/^(.*)\:\d+/)[1] }
```
grosser: we can simplify this to: `line[/^(.*)\:\d+/, 1]`

grosser:
can you reproduce a trace without the `in` ?

keepcosmos:
@grosser 
backtrace can override. 
Actually, I found overrided `#backtrace` method in 126 line of https://github.com/sass/sass/blob/stable/lib/sass/error.rb

Ruby 1.9 document said, 
`The backtrace is an array of strings, each containing either “filename:lineNo: in method”‘ or “filename:lineNo.”`
http://ruby-doc.org/core-1.9.3/Exception.html#method-i-backtrace
We can not sure bracktrace strings match with `/^(.*)\:\d+\:in /`

grosser:
thx, 1.6.1 is out

keepcosmos:
@grosser  Thanks :heart:

PR #472, opened 2016-02-26
Set GEM_PATH as string as it does not accept arrays in RubyGems 2.6.0+

morgoth:
Support for gem path arrays was removed in https://github.com/rubygems/rubygems/commit/27a436e91f1d83dc4a23168851d4d29a73fde752

It fixes https://github.com/rubygems/rubygems/issues/1511

### DIFF
--- a/lib/spring/client/binstub.rb
+++ b/lib/spring/client/binstub.rb
@@ -38,7 +38,7 @@ unless defined?(Spring)
   require 'bundler'
 
   if (match = Bundler.default_lockfile.read.match(/^GEM$.*?^    (?:  )*spring \((.*?)\)$.*?^$/m))
-    Gem.paths = { 'GEM_PATH' => [Bundler.bundle_path.to_s, *Gem.path].uniq }
+    Gem.paths = { 'GEM_PATH' => [Bundler.bundle_path.to_s, *Gem.path].uniq.join(Gem.path_separator) }
     gem 'spring', match[1]
     require 'spring/binstub'
   end


morgoth:
@rafaelfranca Even when one update spring gem, he still need to run `spring binstub --all` to actually fix it - I think issue trackers all over the rails will be bloated by this unfortunately...

rafaelfranca:
Yes. It already started :cry: https://github.com/rails/rails/issues/23907. I'm going to release it right now.

rafaelfranca:
Release 1.6.4

jonleighton:
Thanks both. I've [made a plea](https://github.com/rubygems/rubygems/pull/1094#issuecomment-189395603) to see if we can try to avoid this breakage on the Rubygems side too, since I think it would be vastly preferable to have a deprecation period for the functionality that accepts an array...

lib/spring/client/binstub.rb
```diff
@@ -38,7 +38,7 @@ class Binstub < Command
   require 'bundler'
 
   if (match = Bundler.default_lockfile.read.match(/^GEM$.*?^    (?:  )*spring \((.*?)\)$.*?^$/m))
-    Gem.paths = { 'GEM_PATH' => [Bundler.bundle_path.to_s, *Gem.path].uniq }
+    Gem.paths = { 'GEM_PATH' => [Bundler.bundle_path.to_s, *Gem.path].uniq.join(Gem.path_separator) }
```
jeremy: Apparently `Gem.path_separator` might not be a string: https://github.com/rubygems/rubygems/pull/1476

jeremy: We could work ourselves out of RG internals a bit by using
```ruby
Gem.use_paths Gem.dir, Bundler.bundle_path.to_s, *Gem.path
```

morgoth: Right, I was following from https://github.com/rubygems/rubygems/commit/27a436e91f1d83dc4a23168851d4d29a73fde752

but later it was indeed changed in https://github.com/rubygems/rubygems/commit/544d6fb3475b121a52eec79ab3df8877a2285387

Using `Gem.use_paths` seems like a best way, per @tenderlove comment in https://github.com/rubygems/rubygems/pull/1094#issuecomment-65871904

I'll create PR

PR #473, opened 2016-02-26
Use Gem.use_paths method for setting paths

morgoth:
It fixes case for JRuby which returns a regexp from Gem.path_separator,
but also improves code by using built in method instead of passing string that later is changed to array again.

It follows discussion from https://github.com/rails/spring/pull/472/files#r54284442

### DIFF
--- a/lib/spring/client/binstub.rb
+++ b/lib/spring/client/binstub.rb
@@ -38,7 +38,7 @@ unless defined?(Spring)
   require 'bundler'
 
   if (match = Bundler.default_lockfile.read.match(/^GEM$.*?^    (?:  )*spring \((.*?)\)$.*?^$/m))
-    Gem.paths = { 'GEM_PATH' => [Bundler.bundle_path.to_s, *Gem.path].uniq.join(Gem.path_separator) }
+    Gem.use_paths Gem.dir, Bundler.bundle_path.to_s, *Gem.path
     gem 'spring', match[1]
     require 'spring/binstub'
   end


grosser:
:+1: 

@jeremy double check plz

jeremy:
lgtm /cc @tenderlove

morgoth:
@tenderlove Do you think this solution is good to go?

germanolleunlp:
:+1:

arthurnn:
This happened to me after I upgraded to rubygem 2.6.6. Before I always at 2.4.5.1 and everything was fine.

lib/spring/client/binstub.rb
```diff
@@ -38,7 +38,7 @@ class Binstub < Command
   require 'bundler'
 
   if (match = Bundler.default_lockfile.read.match(/^GEM$.*?^    (?:  )*spring \((.*?)\)$.*?^$/m))
-    Gem.paths = { 'GEM_PATH' => [Bundler.bundle_path.to_s, *Gem.path].uniq.join(Gem.path_separator) }
+    Gem.use_paths Gem.dir, Bundler.bundle_path.to_s, *Gem.path
```
arthurnn: I this going to work on old versions of rubygems?

arthurnn: I see the method is there for a while indeed.

arthurnn:
double checked and it works fine.
Now everyone that has the problem will have the re-generate their binstubs, is that the solution?

PR #474, opened 2016-03-02
Use the Rails 5 reloader API if available

jeremy:
Maintains compat with older Rails as well, but uses the new reloader API if it's available, avoiding a deprecation warning on Rails 5.0 and breakage on Rails 5.1.


rafaelfranca:
:shipit:

Issue #466, opened 2016-01-13
Some certain application fails to reload


tycooon:
So i have 2 apps, one of them restarts just fine:

```bash
$ spring status
Spring is running:

74825 spring server | app1 | started 4 mins ago
75216 spring app    | app1 | started 3 mins ago | development mode

$ touch Gemfile.lock

# Run rails c in other tab – it works just fine but takes some time because app is reloaded.

$ spring status
Spring is running:

74825 spring server | app1 | started 5 mins ago
75498 spring app    | app1 | started 14 secs ago | development mode
```

Bad the other one just hangs when i touch the gemfile:

```bash
bash-3.2$ spring status
Spring is running:

75643 spring server | app2 | started 14 secs ago
75664 spring app    | app2 | started 14 secs ago | development mode

$ touch Gemfile.lock

# Try to run rails c in other tab – it hangs forever

$ spring status
Spring is running:

75643 spring server | app2 | started 1 min ago
75696 (ruby) # this process is dead
```

Both applications use spring 1.6.0 and ruby 2.3.0. Is there a way for me to debug why does the second application hang? It seems to me that the problem is somewhere in the app code, but I have no idea how to debug this.

tycooon:
P.S. All I get in spring log after touching gemfile is this: 

```
[2016-01-13 18:39:19 +0300] [76550] [application:development] preloading app
[2016-01-13 18:39:27 +0300] [76550] [application:development] initialized -> running
/Users/tycooon/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/spring-1.6.0/lib/spring/application.rb:286:in `read': stream closed (IOError)
	from /Users/tycooon/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/spring-1.6.0/lib/spring/application.rb:286:in `block (2 levels) in with_pty'`
```

76550 is the actual process that hangs.

ayufan:
Check this out: https://github.com/rails/spring/pull/484

tycooon:
It turns out the problem was in this initializer in my app:

```ruby
unless Rails.env.production?
  Thread.abort_on_exception = true
end
```

I just removed it and reloading works fine.

ayufan:
Yes, it's. My MR tries to fix that problem.

PR #484, opened 2016-06-20
Configure threads to not abort on exceptions

ayufan:
This solves long outstanding issue: https://github.com/rails/spring/issues/396.

This happens, because some applications or gems do overwrite global setting of `Thread.abort_on_exception=` which is by default set to `false`.
This leads to application process to exit and making the spring unable to recover from that.

Examples of such: 
- https://github.com/alexdalitz/dnsruby/blob/master/lib/dnsruby/select_thread.rb#L27
- https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/thread_safe/synchronized_delegator.rb#L26

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.7.2
+
+* Use `Spring.failsafe_thread` to prevent threads from aborting process due to `Thread.abort_on_exception` when set to `true`
+
 ## 1.7.1
 
 * Specify absolute path to spring binfile when starting the server
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -291,12 +291,7 @@ module Spring
     def with_pty
       PTY.open do |master, slave|
         [STDOUT, STDERR, STDIN].each { |s| s.reopen slave }
-        Thread.new {
-          begin
-            master.read
-          rescue
-          end
-        }
+        Thread.new { master.read }
         yield
         reset_streams
       end
@@ -311,7 +306,7 @@ module Spring
       @mutex.synchronize { @waiting << pid }
 
       # Wait in a separate thread so we can run multiple commands at once
-      Thread.new {
+      Spring.failsafe_thread {
         begin
           _, status = Process.wait2 pid
           log "#{pid} exited with #{status.exitstatus}"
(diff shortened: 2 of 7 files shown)


jonleighton:
This seems good, thanks. Some comments:

* Please add to the changelog file
* I suggest extracting all the repetitive code into a `Spring.failsafe_thread { ... }` method

ayufan:
Updated. Can you review @jonleighton?

tycooon:
I tried this fork in my app (see #466), but no luck. Actually, this version just doesn't work: it tries to reload app right after i run `rspec` (even though i didn't change any files) and hangs the same way it does in original version.

ayufan:
It seems strange, because your problem seems very similar to mine. Can you post maybe more logs? Also can you post what gems are you using?

tycooon:
Here is the log: https://gist.github.com/tycooon/61a711940539643bb17b9151a4cffe7d
I stopped spring and then started/closed rails console several times (it hanged on the 3rd attempt).

ayufan:
@tycooon Are you sure that you are using my changes?

tycooon:
Yes, I've put `gem "spring", github: "ayufan/spring"` in my `Gemfile` and run `bundle`. And spring's behavior actually changed, it doesn't work.

jonleighton:
> Yes, I've put gem "spring", github: "ayufan/spring" in my Gemfile and run bundle. And spring's behavior actually changed, it doesn't work.

This is not a supported way of using spring. If you want to test this branch, you'll need to check it out and build and install the gem manually.

tycooon:
OK, now I made this steps:

`git clone https://github.com/ayufan/spring`
`cd spring && gem build spring.gemspec`
`gem uninstall spring`
`gem install --local ./spring-1.7.1.gem`

Now it doesn't die when I run `rails c` several times but still hangs after trying to reload app after `touch Gemfile` (the same behavior as original 1.7.1 version). Here is the log: https://gist.github.com/tycooon/aa8cba7947563d52bb113304216ef38c

ayufan:
@tycooon Strange. It works without any problems for me now :)

Maybe you can post your gems?

jonleighton:
I've released 1.7.2 with this fix, so you can try that. Thanks!

PR #487, opened 2016-07-28
Support AcitveSupport 5.0.0 as a dependency.

junaruga:
Hi,
Is it possible to support activesupport 5.0.0 too as a dependency?
Thanks.

### DIFF
--- a/spring.gemspec
+++ b/spring.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |gem|
   gem.files         = Dir["LICENSE.txt", "README.md", "lib/**/*", "bin/*"]
   gem.executables   = gem.files.grep(%r{^bin/}).map{ |f| File.basename(f) }
 
-  gem.add_development_dependency 'activesupport', '~> 4.2.0'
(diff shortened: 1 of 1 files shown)


jonleighton:
Hi @junaruga, this seems fine, but why is it needed? AS is only a development dependency, not a proper dependency...

jonleighton:
(Actually, the tests now fail for Ruby versions <= 2.2, so that will need to be dealt with too.)

jeremy:
Supported since 9c03668d187782759bef1e9cba38d2d3733ad34b

junaruga:
Hi @jonleighton , I am so sorry for my late replying.
Actually I missed the notification.

> Hi @junaruga, this seems fine, but why is it needed? AS is only a development dependency, not a proper dependency...

The reason is that I am managing spring's RPM package on Fedora Project.
And a RPM set of a version's Fedora OS basically does not manage 2 kind of RPM versions for a package.
And only AS 5.0.0.1 is used for the RPM set.
So, I wanted to use it for the spring dependency.

PR #497, opened 2016-10-12
Kill PTY I/O thread when we are done with it

misterbyrne:
Sometimes the PTY slave can be closed, causing an unhandled exception in
the thread which is calling `master.read` (socket closed).

This change kills the thread before this happens.

This solves https://github.com/rails/spring/issues/396 for the test case included
 (since this exception was causing Spring to hang)

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -295,10 +295,11 @@ module Spring
     def with_pty
       PTY.open do |master, slave|
         [STDOUT, STDERR, STDIN].each { |s| s.reopen slave }
-        Thread.new { master.read }
+        reader_thread = Spring.failsafe_thread { master.read }
         begin
           yield
         ensure
+          reader_thread.kill
           reset_streams
         end
       end
--- a/lib/spring/test/acceptance_test.rb
+++ b/lib/spring/test/acceptance_test.rb
@@ -172,6 +172,14 @@ module Spring
         assert_success app.spring_test_command
       end
 
+      test "app gets reloaded even with abort_on_exception=true" do
+        assert_success app.spring_test_command
+        File.write(app.path("config/initializers/thread_config.rb"), "Thread.abort_on_exception = true")
+
+        app.await_reload
+        assert_success app.spring_test_command
+      end
+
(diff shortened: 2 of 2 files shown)


misterbyrne:
I haven't been able to create a repeatable test case for this problem without modifying the application code.

However, I have managed to see the exception occur with the following setup:

1. Create a new Rails application (e.g. `rails new spring-test`)
2. Modify the `Thread.new` block in the `with_pty` method of the `lib/spring/application.rb` file of the spring gem to rescue exceptions, e.g.

`vim $(bundle show spring)/lib/spring/application.rb`

...

```rb
        Thread.new do
          begin
            master.read
          rescue => e
            log e
          end
        end
```

3. In one terminal window enter `spring stop; spring start;`
4. In another terminal window

```bash
rails runner 'puts 1 + 1'
touch config/application.rb
```

You'll see a `socket closed` entry logged by the spring server. It won't cause it to hang though, because of the `rescue`.

misterbyrne:
I should probably also add that completely removing the `with_pty` wrapping the `preload` call also fixes the problem of hanging. I assume that it was introduced for a good reason, but wouldn't it be strange for a Rails app to require a PTY during boot? (would a PTY be available when booted from Passenger / pow / etc.?)

lib/spring/application.rb
```diff
@@ -295,10 +295,11 @@ def print_exception(stream, error)
     def with_pty
       PTY.open do |master, slave|
         [STDOUT, STDERR, STDIN].each { |s| s.reopen slave }
-        Thread.new { master.read }
+        reader_thread = Spring.failsafe_thread { master.read }
```
misterbyrne: I've switched to using `Spring.failsafe_thread` here for consistency

misterbyrne:
@jonleighton Does this look okay to you?

jonleighton:
Sorry, I haven't got round to reviewing it yet, but it's on my list and hasn't been forgotten.

mikesea:
@jonleighton could you please release a new version to rubygems? i haven't had any issues with spring hanging since using this version

jonleighton:
@mikesea done

PR #500, opened 2016-11-23
Removing dots in all paragraphs' ends

tjai89:
I have removed all the dots to make the text more balanced. In some places the paragraphs contain dots at the end, in other - do not. I think it's better to make all of them end the same way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,14 +28,14 @@
   process in the foreground, which logging to stdout. This will be
   useful to those who want to run spring more explicitly, but the real
   impetus was to enable running a spring server inside a Docker
-  container.
+  container
 * Numerous other tweaks to better support running Spring inside
   containers (see
   https://github.com/jonleighton/spring-docker-example)
 
 ## 1.6.4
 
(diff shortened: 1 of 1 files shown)


grosser:
👍 I don't like them either ... but there might be some rails guidelines I don't know about ... @rafaelfranca ?

kaspth:
Our API documentation guide says to follow regular punctuation rules, including periods at the end of sentences.

For instance, Active Record's CHANGELOG does so: https://github.com/rails/rails/blob/master/activerecord/CHANGELOG.md

So the inconsistency is the missing periods, not the present ones 😉

tjai89:
So, what is the best option? Remove all dots or add them?

matthewd:
I agree that omitting them can look nice for single-sentence entries. But if there are two sentences, omitting the period on the second just looks weird

Personally, I don't think it's worth bothering to change either way.

tjai89:
I do not want to bother anyone. I will do it myself. Just decide which option is better: leaving dots or removing them, I will do the rest.

kaspth:
Agree with @matthewd, let's just not bother. Thanks!

Issue #330, opened 2014-07-15
Running Spring as a background process


AhmedAttyah:
### Problem:

When we tried to start spring as background process on CircleCI we got this error:

``` bash
/home/ubuntu/fundbase/vendor/bundle/ruby/1.9.1/bundler/gems/spring-5b27cda52354/lib/spring/sid.rb:41:in `getpgid': No such process (Errno::ESRCH)
    from /home/ubuntu/fundbase/vendor/bundle/ruby/1.9.1/bundler/gems/spring-5b27cda52354/lib/spring/sid.rb:41:in `pgid'
    from /home/ubuntu/fundbase/vendor/bundle/ruby/1.9.1/bundler/gems/spring-5b27cda52354/lib/spring/server.rb:78:in `set_pgid'
```

The reason behind this error according to https://github.com/rails/spring/issues/231#issuecomment-31556537  is that CircleCI runs a background process with a command equivalent to `nohup bash -c "./daemon &"` and since there is no associated terminal `Process.getsid` is returning nil or 0.
### Solution:

After some debugging i found that we actually don't need process group ID `pgid` on CircleCI  and that Session ID `sid` is sufficient to keep spring running in background. Thus, i modified `self.pgid` method to be:

``` ruby
def self.pgid
  return sid if ENV['CIRCLECI']
  Process.getpgid(sid)
end
```

So, I've 2 questions:
1- I'm not sure what can go wrong if we relied on `sid` only for background processes?.
2- can we make this configurable ?.

schasse:
+1

oklas:
Hi!
I have same problem on FreeBSD 9.3 (csh)
(Rails 4.2.6 Ruby 2.1.7p400)

when I try to

    bundle exec spring

It crash and result is:

    /home/zYL1WK4y/.bundle/gems/spring-1.6.3/lib/spring/sid.rb:40:in `getpgid': No such process (Errno::ESRCH)
        from /home/zYL1WK4y/.bundle/gems/spring-1.6.3/lib/spring/sid.rb:40:in `pgid'
        from /home/zYL1WK4y/.bundle/gems/spring-1.6.3/lib/spring/server.rb:78:in `set_pgid'
        from /home/zYL1WK4y/.bundle/gems/spring-1.6.3/lib/spring/server.rb:34:in `boot'
        from /home/zYL1WK4y/.bundle/gems/spring-1.6.3/lib/spring/server.rb:14:in `boot'
        from -e:1:in `<main>'

[...]

PR #504, opened 2016-12-26
fix craches where session pgid is unknown #330

oklas:
This changes make it possible to work on some os.
It contains only pgid result checking and catch exceptions if any.

(more about problem in discussion #330)

### DIFF
--- a/lib/spring/server.rb
+++ b/lib/spring/server.rb
@@ -81,7 +81,12 @@ module Spring
     # This will cause it to be automatically killed once the session
     # ends (i.e. when the user closes their terminal).
     def set_pgid
-      Process.setpgid(0, SID.pgid)
+      pgid = SID.pgid
+      if pgid.nil?
+        log "warn: unable to fetch process group of current session"
+        return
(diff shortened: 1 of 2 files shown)


oklas:
@rafaelfranca this does not broke tests and does not add noew feature - this is only fix. Can you please comment or request anything to step to merge?

oklas:
@rafaelfranca Happy new year!

oklas:
Hi @rafaelfranca

This changes actually contains only checking and catch exceptions to make it worked on some os.

Could you review and merge?

rafaelfranca:
Can you confirm this is still a problem with newest release of spring?

oklas:
I can not even confirm problem with one of previous release 2.1.0 - seems it is worked (it can start and stop without crash). It is strange but I think there is no sense to spend time to inspect. The problem was on Freebsd partially because `ps` does not support `-o sess=`. Current version seems does not use such `ps` output. But I notice another minor problem and submit it here #660.

PR #520, opened 2017-04-21
Update .travis.yml

junaruga:
I updated Rubies to latest version.
I also added `ruby-head` as `allow_failures`.

I think this is useful.
Because we can support the next version as faster.

We can see this kind of logic in `rails/rails`, `rspec` and `cucumber` and etc.
https://github.com/rails/rails/blob/master/.travis.yml
https://github.com/rspec/rspec-core/blob/master/.travis.yml
https://github.com/cucumber/cucumber-ruby/blob/master/.travis.yml

Is it possible to merge?

Thanks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,11 @@ sudo: false
 rvm:
   - 1.9.3 # when removed, get rid of the before_script hack and also the one in application_generator.rb
   - 2.0.0
-  - 2.1.9
-  - 2.2.5
-  - 2.3.1
+  - 2.1.10
+  - 2.2.7
+  - 2.3.4
+  - 2.4.1
+  - ruby-head
 env:
   - RAILS_VERSION="~> 4.2.0"
   - RAILS_VERSION="~> 5.0.0"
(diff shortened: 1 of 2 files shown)


junaruga:
https://travis-ci.org/rails/spring/builds/224401092
Ruby 1.9.3 and 2.0.0 with RAILS_VERSION 4.2 are failed, because the dependency nokogiri requires Ruby version >= 2.1.0.
Ruby 2.1.10 with RAILS_VERSION 5.0 is failed, because the dependency activesupport (~> 5.0.0) depends on ruby (>= 2.2.2).

Maybe it's time to drop those case from `.travis.yml`?

rafaelfranca:
We still need to support both versions.

junaruga:
@rafaelfranca ah sorry I will restore Ruby 1.9.3 and 2.0.0.
So it is ok to that those Rubies are tested with `rake test:unit` only on Travis?

rafaelfranca:
Can't we do the same hack that was done for mime-types for nokogiri?

junaruga:
@rafaelfranca ok let me try it.

junaruga:
@rafaelfranca I fixed the tests by your recommended way, and all the test was passed.
The modification of `lib/spring/test/application_generator.rb` is because of below `nokogiri` install error during acceptance test.
https://travis-ci.org/rails/spring/builds/224469456

Can you merge this PR?
Thanks.

junaruga:
ping someone, could you merge it?

jonleighton:
Thanks

PR #522, opened 2017-05-18
Gracefully deal with dangling symlinks

jeremy:
Gracefully deal with dangling symlinks
* Don't add files that are dangling symlinks
* When watching a directory, ignore dangling symlinks rather than treating them as forever-stale

Add a watcher logging facility so such things can be diagnosed and debugged.

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -66,7 +66,17 @@ module Spring
 
     def start_watcher
       @watcher = Spring.watcher
-      @watcher.on_stale { state! :watcher_stale }
+
+      @watcher.on_stale do
+        state! :watcher_stale
+      end
+
+      if @watcher.respond_to? :on_debug
+        @watcher.on_debug do |message|
+          spring_env.log "[watcher:#{app_env}] #{message}"
+        end
+      end
+
       @watcher.start
     end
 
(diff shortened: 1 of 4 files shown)


lib/spring/application.rb
```diff
@@ -67,6 +67,11 @@ def initialized?
     def start_watcher
       @watcher = Spring.watcher
       @watcher.on_stale { state! :watcher_stale }
+
+      @watcher.on_debug do |message|
+        spring_env.log "[watcher:#{app_env}] #{message}"
+      end if @watcher.respond_to?(:on_debug)
```
jonleighton: Why is the conditional needed?

lib/spring/watcher/abstract.rb
```diff
@@ -36,14 +48,30 @@ def add(*items)
           end
         end
 
-        items = items.select(&:exist?)
+        items = items.select do |item|
+          if item.symlink?
+            item.readlink.exist?.tap do |exists|
+              debug do
+                "add: ignoring dangling symlink: #{item.inspect} -> #{item.readlink.inspect}"
+              end if !exists
```
jonleighton: Not really a fan of this style of postfix conditional on a multi-line block personally (I think it's harder to spot what the conditional is doing...)

lib/spring/application.rb
jeremy: Only the poller watcher implements it. Calling it on the listen watcher would bomb.

jonleighton: But it's defined in the abstract class which the listen watcher inherits from?

PR #523, opened 2017-05-18
Watcher: No need to keep polling mtimes if already stale

jeremy:
Since `#mark_stale` only informs listeners of staleness if not already stale, we don't need to keep polling for mtime when we're already stale. Just stop the poller thread entirely.

That saves significant CPU/IO (i.e. laptop battery) if you have, say, an IRB console sitting open.

### DIFF
--- a/lib/spring/watcher/polling.rb
+++ b/lib/spring/watcher/polling.rb
@@ -33,7 +33,7 @@ module Spring
             Thread.current.abort_on_exception = true
 
             begin
-              loop do
+              until stale?
                 Kernel.sleep latency
                 check_stale
               end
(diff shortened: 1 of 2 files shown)


jonleighton:
What's the context for this change? What's the problem that the existing implementation causes?

jeremy:
Not a problem in the sense that anything is broken. Once we're marked as stale, any further checks will just mark as stale again, which no-ops if we're already stale. So: once we detect mtimes have changed, we needn't detect subsequent changes.

That saves significant CPU/IO (i.e. laptop battery) if you have, say, an IRB console sitting open.

Context is #522. I had added logging to see what the watcher is doing, and wow is it doing a ton of costly computed-mtime gathering that it just drops on the floor… since it's already stale.

jonleighton:
> if you have, say, an IRB console sitting open.

Aha, that's the use case I was not thinking about. Makes sense :shipit:

PR #534, opened 2017-08-08
Fix binstubs not being replaced when their quoting style was changed

yhirano55:
I found a **bug**: _The same code is duplicatedly generated_.

# Background

I've run `rubocop -a` after `bundle exec spring binstub --all`.
And it changed to double quotes from single quotes in binstubs:

```ruby
#!/usr/bin/env ruby
begin
  load File.expand_path("../spring", __FILE__)
rescue LoadError => e
  raise unless e.message.include?("spring")
end
APP_PATH = File.expand_path("../config/application", __dir__)
require_relative "../config/boot"
require "rails/commands"
```

Then I've run `bundle exec spring binstub --all` again.

# Before

```diff
$ bundle exec spring binstub --all
* bin/rake: spring inserted # <- Oops
* bin/rails: spring inserted # <- Oops

$ git diff
diff --git a/bin/rails b/bin/rails
index 40d2d5e..7ee6e82 100755
--- a/bin/rails
+++ b/bin/rails
@@ -1,5 +1,10 @@
 #!/usr/bin/env ruby
 begin
+  load File.expand_path('../spring', __FILE__)
+rescue LoadError => e
+  raise unless e.message.include?('spring')
+end
+begin
   load File.expand_path("../spring", __FILE__)
 rescue LoadError => e
   raise unless e.message.include?("spring")
diff --git a/bin/rake b/bin/rake
index 3df5805..72f4b15 100755
--- a/bin/rake
+++ b/bin/rake
@@ -1,5 +1,10 @@
 #!/usr/bin/env ruby
 begin
+  load File.expand_path('../spring', __FILE__)
+rescue LoadError => e
+  raise unless e.message.include?('spring')
+end
+begin
   load File.expand_path("../spring", __FILE__)
 rescue LoadError => e
   raise unless e.message.include?("spring")
diff --git a/bin/spring b/bin/spring
index b44ad1a..fb2ec2e 100755
--- a/bin/spring
+++ b/bin/spring
@@ -4,14 +4,14 @@
 # It gets overwritten when you run the `spring binstub` command.

 unless defined?(Spring)
-  require "rubygems"
-  require "bundler"
+  require 'rubygems'
+  require 'bundler'

   lockfile = Bundler::LockfileParser.new(Bundler.default_lockfile.read)
   spring = lockfile.specs.detect { |spec| spec.name == "spring" }
   if spring
     Gem.use_paths Gem.dir, Bundler.bundle_path.to_s, *Gem.path
-    gem "spring", spring.version
[...]

PR #535, opened 2017-08-08
WIP: Prefer __dir__ over __FILE__ in File.expand_path

yhirano55:
For performance.


yhirano55:
Related #534

yhirano55:
I've closed this because it seems be not worth for us. We don't hope that old formats are increased.

PR #540, opened 2017-09-13
Use single quotes throughout `lib/spring/client/binstub.rb`

SpringMT:
Fix to use single quotes throughout `lib/spring/client/binstub.rb`


rafaelfranca:
Merged in ec03cc7

PR #546, opened 2017-11-08
Use Bundler.with_original_env so that BUNDLER_APP_CONFIG is honored

### LINKED ISSUE #545: Spring binstub fails when BUNDLE_APP_CONFIG is set
### Summary

The `BUNDLE_APP_CONFIG` environment variable, if set, tells bundler to use a different directory to store its config file (instead of the default `./.bundle`). Spring does not pass this variable to its child processes. Therefore when a Spring child process requires `bundler/setup`, bundler does not look in the correct location for its config. This can lead to fatal errors when using spring binstubs.

### Steps to reproduce

The bug can be reproduced by setting `BUNDLE_APP_CONFIG` and installing gems using the `--path` option.

We'll use Rails since it automatically sets up binstubs for us.

```
$ gem install rails -v 5.1.4
$ rails _5.1.4_ new binstub-error --no-rc
```

Setting `BUNDLE_APP_CONFIG` and then installing the bundle into `vendor/bundle` means that from hereon out `bundler/setup` will only work properly if `BUNDLE_APP_CONFIG` is honored.

```
$ mkdir -p /tmp/bundle-config-dir
$ export BUNDLE_APP_CONFIG=/tmp/bundle-config-dir
$ cd binstub-error
$ bundle install --path=vendor/bundle
```

Remove the system version of `rake` so that the only way to get it is via `vendor/bundle`:

```
$ gem uninstall rake --all
```

Notice that now the binstubs fails:

```
$ bin/rake --version
Could not find rake-12.2.1 in any of the sources
Run `bundle install` to install missing gems.
```

However `bundle exec` works fine:

```
$ bundle exec rake --version
rake, version 12.2.1
```

If we put the bundler config in the default location and unset `BUNDLE_APP_CONFIG`, the binstub now works:

```
$ cp -R $BUNDLE_APP_CONFIG .bundle
$ unset BUNDLE_APP_CONFIG
$ bin/rake --version
Running via Spring preloader in process 96969
rake, version 12.2.1
```

### Workaround

Do not use `BUNDLE_APP_CONFIG`.

Note that Circle CI sets the `BUNDLE_APP_CONFIG` variable by default in its `ruby:2.4.2-node-browsers` docker image, so many developers could run into this bug.

mattbrictson:
Fixes #545.

Bundler's `with_clean_env` completely erases all Bundler-related environment variables, even if those variables existed prior to loading Bundler. That means that custom Bundler settings that we legitimately want to pass down to child processes - namely, `BUNDLE_APP_CONFIG` - are erased, leading to errors when the child process loads `bundler/setup`.

Note that `clean_env` (used by `with_clean_env`) is actually deprecated by Bundler. The method recommended by Bundler going forward, and the one used in this commit, is `with_original_env`. This clears any variables that were set by Bundler itself, but retains any user-supplied settings that existed prior.

This fixes a bug where spring binstubs would fail if `BUNDLE_APP_CONFIG` was set to a custom location (i.e. other than the default `./.bundle`) and `bundle install --path` was used. The binstub would not be able to load the bundle because without `BUNDLE_APP_CONFIG`, it could not know the path where the gems were installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Fix binstubs not being replaced when their quoting style was changed (#534)
 * Preserve comments right after the shebang line which might include magic comments such as `frozen_string_literal: true'
+* Fix binstub failures when Bundler's `BUNDLE_APP_CONFIG` environment variable is present (#545)
 
 ## 2.0.2
 
--- a/lib/spring/application_manager.rb
+++ b/lib/spring/application_manager.rb
@@ -92,7 +92,7 @@ module Spring
     def start_child(preload = false)
       @child, child_socket = UNIXSocket.pair
 
-      Bundler.with_clean_env do
+      Bundler.with_original_env do
         @pid = Process.spawn(
           {
             "RAILS_ENV"           => app_env,


cmwall:
Our builds were failing at the test suite stage in Circle CI because the `rake` gem was not being found in any of our sources.

We updated our Gemfile to use this branch for `spring` and the problem was solved.

Thanks!

The error we received was:

```
#!/bin/bash -eo pipefail
xvfb-run -a \
./bin/rspec --profile 10 \
--format RspecJunitFormatter \
--out ./test-results/rspec.xml \
--format progress \
$(circleci tests glob "spec/**/*_spec.rb" | circleci tests split --split-by=timings)
Could not find rake-12.2.1 in any of the sources
Run `bundle install` to install missing gems.
Exited with code 1
```

mattbrictson:
@rafaelfranca in your opinion is this fix worth pursuing? It seems that at least a handful of other people have found that this branch fixes their problems using spring binstubs on CircleCI.

I can spend some more time on this PR to try and fix the Travis failures, but first I want to make sure it has a good chance of being accepted. Thanks!

anvox:
The same problem to me, CircleCI 2.0. I must disable spring on CI :( https://github.com/rails/spring#temporarily-disabling-spring 
Under company policy, I dont want to put a dev branch gem to app.

sponomarev:
@mattbrictson Do you know what introduced that problem? We haven't experienced this problem about a month ago, but now it's a common problem for local and Docker environment

matiasleidemer:
I have the same issue in a Docker environment. Updating bundler to `1.16.1`causes the error, but after monkeypatching spring to use `Bundler.with_original_env` it works like a charm. Downgrading bundler to `1.15.4` also fixes the issue.

I'm not sure how this change would impact other projects, however. As a side note, we don't have the `BUNDLE_APP_CONFIG` env variable.

rmm5t:
Confirmed that I am experiencing the same underlying problem and that this change fixes it.

What's the likelihood of something like this making it into a new release? I guess I'm not sure I fully understand the benefits of `.with_clean_env` vs `.with_original_env`.

sponomarev:
`with_clean_env ` was legacy and have a lot of caveats - not all the proper ENV variables are cleaned. It will be deprecated in Bundler 2.0. Recent Bundler major version changed the logic a little bit (to correct it), but, unfortunately, Spring relies on that incorrect behaviour in some limited cases. 

As a temporary solution, we set `GEM_HOME` and `BUNDLE_APP_CONFIG` to our own location if we want to use custom `BUNDLE_PATH`.

Not actually related, but could be useful. Our part of Dockerfile

```Dockerfile
ENV GEM_HOME /bundle
ENV BUNDLE_PATH=$GEM_HOME \
  BUNDLE_APP_CONFIG=$BUNDLE_PATH \
  BUNDLE_BIN=$BUNDLE_PATH/bin
ENV PATH /app/bin:$BUNDLE_BIN:$PATH
```

rmm5t:
> `with_clean_env` was legacy and have a lot of caveats - not all the proper ENV variables are cleaned. It will be deprecated in Bundler 2.0

Thanks for that explanation. Assuming this is all true, what's holding up this PR from being applied?
Obviously the build failed, but it looks like the build is failing on the `master` branch for the same reasons. 

Where is help needed to move this forward? I have some incentive and willingness to help here where I can.

matthewd:
What are the version/dependency implications of this change / when did Bundler introduce `with_original_env`?

Is it safe to just switch outright, or might there be a need to support both methods?

rmm5t:
@matthewd [`with_original_env` was introduced 7 years ago](https://github.com/bundler/bundler/commit/5cf3fde78c2660ff8a815f376921438fd641c67d), so sometimes around bundler v1.0.x.

sponomarev:
Here is the proof from Bundler's source code 

https://github.com/bundler/bundler/blob/23dfadcd8d561d48e89c7287cbbdd816e0a10ab8/lib/bundler.rb#L281-L284

pradyumna2905:
Facing the same issue in GitLab CI as well.

Since I'm just starting out with my app and don't have a ton of tests, I'm disabling spring till this gets fixed.

`DISABLE_SPRING=true bin/rails test`

ajacksified:
I wasted my morning trying to figure out why `rails s` worked but `rails g` didn't; it was this issue. Commenting `spring` out of the Gemfile immediately solved my issue - we use a virtualenv for projects in a dev vm, and `.bundle/config` points at `vendor/bundle`.

Can we merge this one-line change that's been open for half a year? It's especially bad because it's in the default Gemfile for new rails projects, and the error is pretty difficult to track down.

rmm5t:
Thank you, @matthewd!!

Will there be a new versioned release soon, or should we plan to work against the master branch for the near term?

matthewd:
I'd like to get master's CI properly green before we cut a release.

For this change: I'm happy that only Bundler >= 1.0 < 1.1 had `with_clean_env` and not `with_original_env`, and that substantially predates when we started using it.

mtsmfm:
@matthewd Can I ask you to cut new release?

welkie:
Just wanted to mention that I needed to resort to using the `DISABLE_SPRING=true bundle exec rails test` command in my CircleCI config for it to work, despite this branch being merged in. Without disabling Spring, it did not find the Rake gem:

```
#!/bin/bash -eo pipefail
bundle exec rails test
Could not find rake-12.3.1 in any of the sources
Run `bundle install` to install missing gems.
Exited with code 1
```

welkie:
@davidstosik Awesome. I've been out of the loop with Rails for like 2 years but I wanted to update a project's dependencies and throw it on CircleCI. So this threw me off. Just wanted to comment so if anyone else comes here from Googling, they know a fix is still needed.

@literally_everyone Ty for the fix. :)

ypresto:
FYI:
binstub (`bin/spring`) won't work with `gem 'spring', github: 'rails/spring'`.
It assumes spring gem to be installed on standard gem path, not bundler's git repo directory.

ypresto:
Is there ETA to release this? I got the same issue on another project :cry:

rmm5t:
@matthewd I recognize and appreciate that you're waiting to get `master` fully green before pushing out a new release, but spring is kind of dead in the water without this fix released. It's not really possible to use the `master` branch as a gem dependency, because the spring binstub doesn't know how to find the path to spring if it's installed via a bundler `gem "spring", github: "rails/spring"` declaration, because the spring binstub only searches in `Bundler.bundle_path` and not also `Bundler.install_path`

https://github.com/rails/spring/blob/6f1203afd567fcebaf755619e90690e15396825c/lib/spring/client/binstub.rb#L39-L41

bobbytables:
Can we please release this?

ypresto:
Today I faced to this issue, again.
I know you are busy, but, just a release, not a review or writing code.
Please, please release it..! It has been 1.5 year after the PR..!! :joy: :joy: :pray: :pray:

ypresto:
Please mention me if there is any blocker for release and perhaps I can help solving it.

rafaelfranca:
The CI is very broken so we can't release in that way. This PR may be the cause why it is broken.

rmm5t:
> This PR may be the cause why it is broken.

@rafaelfranca The build in the master branch has been green multiple times after this PR was merged.  And it was red multiple times before this PR was merged.

rafaelfranca:
Yes, I confirmed this PR is not the cause

davidstosik:
> Yes, I confirmed this PR is not the cause

If it is not, is there any chance we get a release soon? 🤞

y-yagi:
@rafaelfranca Do you have a plan to release a new gem? I think that this fix is also related to https://github.com/bundler/bundler/issues/7184. If need to fix CI, I will do it.

ypresto:
`fix-ci` branch looks already passing CI..!

rmacklin:
Since no one mentioned it explicitly in this thread, I'll note that [`v2.1.0`](https://github.com/rails/spring/releases/tag/v2.1.0) was eventually released (v2.1.0) which includes this change.

PR #548, opened 2017-12-04
Fix hangs when deleting / recreating project root with the same name

### LINKED ISSUE #247: Spring hangs if the project directory gets removed.
In a situation where spring server has started but then the project directory is deleted, `spring` throws an error:

``` sh
/Users/greg/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/spring-1.1.0/lib/spring/configuration.rb:37:in `pwd': No such file or directory - getcwd (Errno::ENOENT)
    from /Users/greg/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/spring-1.1.0/lib/spring/configuration.rb:37:in `project_root_path'
    from /Users/greg/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/spring-1.1.0/lib/spring/configuration.rb:28:in `application_root_path'
    from /Users/greg/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/spring-1.1.0/lib/spring/application.rb:241:in `loaded_application_features'
    from /Users/greg/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/spring-1.1.0/lib/spring/application.rb:102:in `ensure in preload'
    from /Users/greg/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/spring-1.1.0/lib/spring/application.rb:107:in `preload'
    from /Users/greg/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/spring-1.1.0/lib/spring/application/boot.rb:15:in `<top (required)>'
    from /Users/greg/.rbenv/versions/2.0.0-p247/lib/ruby/site_ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /Users/greg/.rbenv/versions/2.0.0-p247/lib/ruby/site_ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from -e:1:in `<main>'
```

caused by `Dir.pwd` from [here](https://github.com/rails/spring/blob/master/lib/spring/configuration.rb#L37).

One solution is that spring could shut down upon encountering that error.

This can be replicated for version `1.1.0` by following these steps (app setup):

``` sh
rails new spring-test -B
cd spring-test
vi Gemfile

# add `gem 'spring'` to the Gemfile and save.

bundle install
bundle exec spring binstub --all
rails g model person
```

By now spring server should be running. Now if we remove the app directory

``` sh
cd .. && rm -rf spring-test
```

and try to re-run the steps from above (the app setup steps), the last command `rails g ...` will hang.

misterbyrne:
Fix for issue detailed in https://github.com/rails/spring/issues/396#issuecomment-345387012

If we have started a spring server, but the project root is deleted, the server can't successfully boot a child process. This is not recoverable. Even when the project is recreated, clients can also no longer communicate with the server because the socket path is derived from the project root (`Dir.pwd`) which for the server process no longer exists.

The proposed fix is to detect whenever a child cannot be started within a specific time frame - for whatever reason - rather than waiting indefinitely for the child to communicate with the `ApplicationManager` instance. This raises an error. If any error occurs when attempting to create a child process, we just abort - which shuts down the server gracefully.

This might not be a terribly common scenario, but I guess it can happen in real life - especially when people create throwaway projects to test stuff, or check out open source projects temporarily.

Fixes #247

### DIFF
--- a/lib/spring/application_manager.rb
+++ b/lib/spring/application_manager.rb
@@ -34,7 +34,11 @@ module Spring
     end
 
     def alive?
-      @pid
+      return false if @pid.nil?
+      Process.getpgid(@pid)
+      true
+    rescue Errno::ESRCH
+      false
     end
 
     def with_child
--- a/lib/spring/application_manager.rb
+++ b/lib/spring/application_manager.rb
@@ -118,8 +118,18 @@ module Spring
         )
       end
 
-      start_wait_thread(pid, child) if child.gets
+      wait_for_child_to_boot
+      start_wait_thread(pid, child)
       child_socket.close
+    end
+
+    def wait_for_child_to_boot
+      timeout = 1
+      loop do
+        break if IO.select([child], nil, nil, timeout)
+        raise "child has exited unexpectedly" unless alive?
+      end
+      child.gets
     end
 
     def start_wait_thread(pid, child)
--- a/lib/spring/application_manager.rb
+++ b/lib/spring/application_manager.rb
@@ -120,6 +120,10 @@ module Spring
 
       wait_for_child_to_boot
       start_wait_thread(pid, child)
+    rescue => e
+      log "error while starting child: #{e.message}"
+      abort("error while starting child: #{e.message}")
+    ensure
       child_socket.close
     end
 
--- a/test/support/acceptance_test.rb
+++ b/test/support/acceptance_test.rb
@@ -248,6 +248,23 @@ module Spring
         assert_success app.spring_test_command
       end
 
+      test "deleting the app root should kill the server" do
+        app.run app.spring_test_command
+        assert spring_env.server_running?, "The server should be running but it isn't"
+        FileUtils.rm_rf(app.root)
+        sleep 2 # wait for the watcher to notice and react
+        assert !spring_env.server_running?, "The server should not be running but it is"
+      end
+
+      test "server restarts when the app root is deleted and recreated" do
+        assert_success app.spring_test_command
+        FileUtils.rm_rf(app.root)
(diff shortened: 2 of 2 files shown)


lib/spring/application_manager.rb
```diff
@@ -32,7 +32,11 @@ def restart
     end
 
     def alive?
-      @pid
+      return false if @pid.nil?
+      Process.getpgid(@pid)
```
misterbyrne: This seems to be a reasonable way of determining whether a process with a certain pid exists.

lib/spring/application_manager.rb
```diff
@@ -109,10 +113,24 @@ def start_child(preload = false)
...
+      log "error while starting child: #{e.message}"
+      abort("error while starting child: #{e.message}")
+    ensure
       child_socket.close
     end
 
+    def wait_for_child_to_boot
+      timeout = 1
+      loop do
+        break if IO.select([child], nil, nil, timeout)
```
misterbyrne: In most cases, the child process quickly `puts` something on the socket to indicate that it's ready. Rather than simply wait forever as before (i.e. `child.gets`) we do the same but with a 1 second timeout.

misterbyrne:
☝️ I'm not sure why CI failed. I tried a noop PR and that also failed :/

rafaelfranca:
CI is failing so I'll to close this PR.

Issue #515, opened 2017-02-27
Need to increase `BOOT_TIMEOUT` or make it configurable


Abrakazoo:
I need to run `DatabaseCleaner.clean_with(:truncation)` before forking so I added it to config.spring.rb. This caused spring server to exit with timeout after 20 seconds. I looked for a way to patch this by extending Spring::Client::Run but to no avail, for some reason. Added the patch to both lib/ and initializers, still not working. 
Forking only to change this constant seems like quite an overkill.
What would you recommend I do in this case? Submit a PR? fork?
Thanks!

jonleighton:
A PR to make this configurable seems reasonable

jackscotti:
"Nice" to hear someone had the same problem! 

Here's the PR: https://github.com/rails/spring/pull/554

Sadly is failing for something that does not seem related to my changes (works in local). @jonleighton do you have any suggestion or advice for me?

rafaelfranca:
`Spring.boot_timeout` was added

PR #554, opened 2018-02-02
Make `BOOT_TIMEOUT` configurable with env var

jackscotti:
👋 

## What:
This PR makes `BOOT_TIMEOUT` configurable thanks to the addition of the `SPRING_BOOT_TIMEOUT` environment  variable.

## Why:
We needed Spring to wait longer than 20 seconds but the time was hardcoded

## Addresses issue: https://github.com/rails/spring/issues/515

** I'd be very happy to amend the copy if not clear enough**


jackscotti:
I'm a bit puzzled by the results on Travis, seems that the failures are due to config issues not related to my changes (?). 🆘

PR #559, opened 2018-05-21
Refer to Spring as a proper noun, with a capital "S"

timrogers:
This updates the documentation and the code itself (particularly the comments and bits of the code that generate help, e.g. when you run `spring --help`) to refer to "Spring" with a capital "S", since it's a name (i.e. a proper noun).

The only places where I've left "spring" with a lower-case "s" is where we're referring to the name of the gem (e.g. in a `Gemfile`) or for code examples (e.g. running `bin/spring`).


timrogers:
I found this while working on #560, and thought it would be nice to be consistent.

timrogers:
Are they due to my change? I had very similar in #560, which only changed the readme. I don’t have these failures locally.

timrogers:
Oh yes! Let me look at that. I couldn’t run the acceptance tests locally.

PR #560, opened 2018-05-21
Document how to check whether Spring is running from within your code

timrogers:
You can check whether Spring is running by checking whether the `Spring` constant is defined.

```ruby
if defined?(Spring)
  # This code will only run if Spring is running.
end
```

This is the approach taken by Spring binary, [`bin/spring`][1].

I've often personally wanted to check whether Spring is running (e.g. when I've been seeing strange behaviour that I suspect is related to class reloading), and this is potentially useful for other gems too (e.g. to warn if environment variables might not be reflected as expected due to running through the Spring preloader).

[1]: https://github.com/rails/spring/blob/master/bin/spring#L3

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,24 @@ $ bin/spring stop
 Spring stopped.
 ```
 
+From within your code, you can check whether Spring is active by checking whether the
+`Spring` constant is defined:
+
+```ruby
+# With Spring running
+defined?(Spring)
+# => "constant"
+
+# With Spring disabled
+defined?(Spring)
+# => nil
+
+# Since a string is truthy and `nil` is falsy, we can use this as our condition
+if defined?(Spring)
+  # Do something
+end
+```
+
 ### Removal
 
 To remove spring:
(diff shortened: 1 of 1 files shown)


grosser:
not that critical of information, how about condensing it into a 1-liner ?

"You can check if spring is loaded with `if defined?(Spring)`"

timrogers:
Sure. Coming right up.

timrogers:
@grosser Oh, and thanks for the very quick review!

grosser:
... also I think it's wrong since `bin/rails`always loads spring so it will always be defined.

timrogers:
@grosser I've made that change.

That's interesting - what you've said doesn't seem to be borne out by my experience, but I could be missing something.

If I run `spring stop`, then `bin/rails c`, `Spring` is defined. If I stop Spring again, and then run `DISABLE_SPRING=true bin/rails c`, then `Spring` isn't defined. What am I missing?

grosser:
ah interesting, might be alright then :)
I just saw that it always loads spring and though it would mean it's always defined ... but apparently it's not since that is handled here https://github.com/rails/spring/blob/master/lib/spring/binstub.rb
👍

grosser:
looks like you broke all the tests :trollface:

PR #569, opened 2018-09-28
Update ruby, rails version for travis builds

lostapathy:
This updates the travis config for the latest ruby and rails versions.

* --skip-listen was added because travis was having intermittent failures with the Guard gem
* --skip-system-test was added because it creates a dependency on capybara in the generated app which does not resolve in a buildable way with ruby versions less than 2.3.

I don't believe either of these changes affect the utility of the test suite.


lostapathy:
The current travis failure seems to be an issue with their build system.  While testing, I saw this build fail this way about 25% of the time for no apparent reason.  It will most likely run if someone with the correct access tells it to retry.

PR #575, opened 2018-11-14
Setup "pgid" for server running in foreground mode

kradima:
It's expected that `pgid` to be setup for "foreground" running server rather than for "background".

So it should be changed to `set_pgid if foreground?`

See discussion https://github.com/rails/spring/commit/e16e085e4e6aada8b32198e98e0c8ae1ab7b9963#r18804964

### DIFF
--- a/lib/spring/server.rb
+++ b/lib/spring/server.rb
@@ -36,7 +36,7 @@ module Spring
       Spring.verify_environment
 
       write_pidfile
-      set_pgid unless foreground?
+      set_pgid if foreground?
       ignore_signals unless foreground?
       set_exit_hook
       set_process_title


grosser:
@jonleighton @develop7 @hjwylde plz review

hjwylde:
LGTM

develop7 (approved):
Looks good

develop7:
:shipit:

jonleighton:
My memory of the spring code is not super fresh, but I think the code is correct as-is and this should not be merged.

The idea of the `set_pgid` method is that it ties the background spring process to the current terminal session. Therefore, when the terminal session ends, the background process also gets killed.

So I think merging this will result in orphaned spring processes accumulating on users' machines.

When running in 'foreground mode', the spring process is never detached. Therefore it will just be up to the parent process to kill it when necessary.

If people want to run spring outside of a terminal session, they should do so using `spring server`, then they can manage the starting/stopping manually as desired.

PR #577, opened 2018-12-01
Set quiet mode from environment

bobf:
If `SPRING_QUIET` is set, do not display message each time a command is
run. This takes precedence over `Spring.quiet`.

Rationale:

I tend to use a `Makefile` for various orchestration commands (such as calling a number of `rake` commands). The tailored output from these commands gets a bit clobbered by Spring's `"Running via Spring preloader in process"` message (often appearing multiple times for one `make` command).

If I can prefix my `make` commands with `SPRING_QUIET=1` then things will work a lot more nicely.


thewoolleyman:
Duplicate of #674 

See comment for details: https://github.com/rails/spring/pull/674#issuecomment-1114492978

PR #584, opened 2019-02-27
Add command line option to allow eager preloading for rails commands

### LINKED ISSUE #568: Eager fork?
When I was using zeus, it forks in advance every thing: so rails g, rails c, rails s, rake should run instantly. Not so much with spring. Everytime there is necessary reload, spring sit in the background doing nothing, and the next invoke of command will have to wait for the fork to happen. I would suggest one launched, spring should actively re-fork itself.

PikachuEXE:
Fixes #568 

Tested with app env `development` only

### DIFF
--- a/lib/spring/server.rb
+++ b/lib/spring/server.rb
@@ -17,15 +17,20 @@ module Spring
     attr_reader :env
 
     def initialize(options = {})
-      @foreground   = options.fetch(:foreground, false)
-      @env          = options[:env] || default_env
-      @applications = Hash.new { |h, k| h[k] = ApplicationManager.new(k, env) }
-      @pidfile      = env.pidfile_path.open('a')
-      @mutex        = Mutex.new
+      @foreground     = options.fetch(:foreground, false)
+      @eager_preload  = options.fetch(:eager_preload, false)
+      @env            = options[:env] || default_env
+      @applications   = Hash.new { |h, k| h[k] = ApplicationManager.new(k, env) }
+      @pidfile        = env.pidfile_path.open('a')
+      @mutex          = Mutex.new
     end
 
     def foreground?
       @foreground
+    end
+
+    def eager_preload?
+      @eager_preload
     end
 
     def log(message)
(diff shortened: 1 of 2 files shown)


lib/spring/server.rb
```diff
@@ -46,6 +51,7 @@ def boot
     def start_server
       server = UNIXServer.open(env.socket_name)
       log "started on #{env.socket_name}"
+      eager_fork! if eager_fork?
```
JuPlutonic: Should we add somth like `…if eager_fork?|| !env.JRuby? ` ?
JRuby didn't implement fork.

PikachuEXE: It's actually calling `ApplicationManager#restart` to start preloading
https://github.com/rails/spring/blob/v2.0.2/lib/spring/application_manager.rb#L29

Maybe it can be renamed to eager_preload

PikachuEXE:
Will update the naming from "fork" to "preload" since ["preload" is the env var set](https://github.com/rails/spring/blob/v2.0.2/lib/spring/application_manager.rb#L101)

PikachuEXE:
Any more comments or changes required?

PikachuEXE:
Using my own branch without issue
(I put `bundle exec spring server --eager-preload` as a container command)

Would be great to have this merged

PikachuEXE:
Any comment for this PR as `2.1.0` released?

PikachuEXE:
Is there anything I should do before this gets merged?

PikachuEXE:
Anyone tested/using this change?

phuongnd08:
I really like this one, but hasn't got decent time to try this. My suggestion is that you can either mention the maintainer guy rafaelfranca, and post some benchmark to help the guys see the benefit of this one. It's funny that we have to spend effort making improvement, then more effort for it to be noticed. Not very encouraging for contribution.

milushov:
@rafaelfranca bump

PikachuEXE:
Any review comment for this PR?

PikachuEXE:
Rebased on latest `master`
Forgot to do so when 2.1.1 released

PikachuEXE:
Unable to use spring on production anymore due to https://github.com/rails/spring/commit/24cc80e256531e62c56a3224981cb8f9d7f56061

PR #589, opened 2019-04-04
Rails 5.0.0 needs sqlite3 1st, rb 2.1.10,2.2.10 needs older i18n/nokogiri

JuPlutonic:
Added if condition $RAILS_VERSION when it equals 5.0.0.
Added if condition travis_ruby 2.1.10 2.2.10 to install i18n 1.5.1
AcceptanceTest#test_custom_bundle_path and all following tests tests didn't pass [eg. see](https://travis-ci.org/rails/spring/jobs/513069834)


JuPlutonic:
@travis please re-run tests👏

PR #593, opened 2019-06-10
Ruby Warnings

amatsuda:
This PR fixes some Ruby warnings.

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -147,7 +147,7 @@ module Spring
       log "got client"
       manager.puts
 
-      stdout, stderr, stdin = streams = 3.times.map { client.recv_io }
+      _stdout, stderr, _stdin = streams = 3.times.map { client.recv_io }
       [STDOUT, STDERR, STDIN].zip(streams).each { |a, b| a.reopen(b) }
 
       preload unless preloaded?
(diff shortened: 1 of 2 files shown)


lib/spring/application.rb
```diff
@@ -147,7 +147,7 @@ def serve(client)
       log "got client"
       manager.puts
 
-      stdout, stderr, stdin = streams = 3.times.map { client.recv_io }
+      _stdout, stderr, _stdin = streams = 3.times.map { client.recv_io }
```
jeremy: Seems we aren't using these local variables at all. Can we just remove them?

lib/spring/application.rb
```diff
@@ -147,7 +147,7 @@ def serve(client)
...
       [STDOUT, STDERR, STDIN].zip(streams).each { |a, b| a.reopen(b) }
```
jeremy: Might be clearer to do this explicitly, even 😅
```diff
-stdout, stderr, stdin = streams = 3.times.map { client.recv_io }
-[STDOUT, STDERR, STDIN].zip(streams).each { |a, b| a.reopen(b) }
+STDOUT.reopen(client.recv_io)
+STDERR.reopen(client.recv_io)
+STDIN.reopen(client.recv_io)
```

lib/spring/application.rb
rafaelfranca: We are using the `stderr` in the `rescue` block

PR #594, opened 2019-06-24
Test with Rails 6.0.0 rc1

robertomiranda:
Rails Release candidate was released almost 2 months ago why not test against that version?


robertomiranda:
cc @rafaelfranca 🙇

PR #602, opened 2019-10-24
CI: Use ruby 2.4.9, 2.5.7, 2.6.5 and rails 6.0.0

mikong:
Bump the Ruby versions used by Travis CI. Also update from Rails 6.0.0.rc1 to 6.0.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,21 @@
 language: ruby
 cache: bundler
 rvm:
-  - 2.4.6
-  - 2.5.5
-  - 2.6.3
+  - 2.4.9
+  - 2.5.7
+  - 2.6.5
   - ruby-head
 env:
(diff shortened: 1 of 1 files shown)


mikong:
The build is failing because it's unable to install rails. sprockets now requires Ruby version >= 2.5.0. Should we drop support for Ruby 2.4.x?

Or use sprockets 3.7.2 instead of 4.0.0?

rafaelfranca:
Lets use sprockets 3.7.

mikong:
Alright. I updated my PR to install sprockets 3.7 under `before_install`. However, that means it's installed for all jobs. Please let me know if sprockets 3.7 should only be installed for Ruby 2.4. I'm not sure what the best way to do that though.

byroot:
Closing as outdated. Thanks for the patch though.

PR #617, opened 2020-06-10
Remove support for old Rails versions

deivid-rodriguez:
I noticed several CI failures on #616. Since all of them happen on Rails versions no longer maintain, I think that's a good sign that it's time to drop support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ rvm:
   - 2.6.3
   - ruby-head
 env:
-  - RAILS_VERSION="~> 5.0.0"
   - RAILS_VERSION="~> 5.1.0"
   - RAILS_VERSION="~> 5.2.0"
   - RAILS_VERSION="~> 6.0.0.rc1"
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ rvm:
   - 2.6.3
   - ruby-head
 env:
-  - RAILS_VERSION="~> 4.2.0"
   - RAILS_VERSION="~> 5.0.0"
   - RAILS_VERSION="~> 5.1.0"
   - RAILS_VERSION="~> 5.2.0"
@@ -15,8 +14,6 @@ matrix:
   exclude:
     - rvm: 2.4.6
       env: RAILS_VERSION="~> 6.0.0.rc1"
-    - rvm: 2.6.3
-      env: RAILS_VERSION="~> 4.2.0"
   allow_failures:
     - rvm: ruby-head
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ rvm:
   - 2.6.3
   - ruby-head
 env:
-  - RAILS_VERSION="~> 5.1.0"
   - RAILS_VERSION="~> 5.2.0"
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
I went crazy and also dropped support for ruby 2.4, which is also unsupported upstream. CI is green now.

Feel free to cherry-pick any of these changes if they are of interest :)

elia:
@guilleiguaran or anyone else from the core team, I think this one should be reviewed to put the CI back on track
@deivid-rodriguez can you add a note on how [the commit you reverted](https://github.com/rails/spring/commit/20d3a5f03e0e758a887cfa95d63c0af185d79341) was breaking the build?

deivid-rodriguez:
Coming here from #629 that just got merged! :tada:

Now, in order to get CI green, I see a couple of ways:

* Drop support for whatever it's failing, since it's unsupported upstream anyways. This is what this PR does and I see it as "the easy way" in terms of implementation but "the hard way" in terms of decision making, since a policy for dropping support for old Rails and Rubies need to be decided.

* Fix things. This is what, at least partially, #630 does. This is harder but less debatable since it doesn't involve dropping support for anything.

Thanks @Sixeight for your work by the way :smiley:.

.travis.yml
```diff
@@ -1,22 +1,13 @@
...
-  - 2.4.6
   - 2.5.5
   - 2.6.3
   - ruby-head
 env:
-  - RAILS_VERSION="~> 4.2.0"
-  - RAILS_VERSION="~> 5.0.0"
-  - RAILS_VERSION="~> 5.1.0"
   - RAILS_VERSION="~> 5.2.0"
   - RAILS_VERSION="~> 6.0.0.rc1"
```
elia, line 9: I think this can be bumped to `~> 6.0.0` now

elia (changes requested):
Thanks for fixing this, I hope the project will be soon back to green ✅ 
Just a nit over not using RC versions, which I've found in the recent past to be quite deceptive.

deivid-rodriguez:
Thanks for the review @elia! I observed the same thing, and also that ruby 2.7 is not currently tested.

I was expecting to get some feedback here and then maybe make those updates separately on a different PR, but I can also add that change here if maintainers prefer. I was just afraid that something might fail and then I have to fix it, then the PR starts growing and I keep going down the rabbit hole. I think dropping support is a good scope for this PR.

deivid-rodriguez:
Hi again maintainers :)!

I was thinking that, given that Rails itself has a quite aggressive policy regarding supported rubies (which I agree with), I think dropping support for ruby 2.4 seems fine? Since Rails 6, only ruby 2.5 and above are supported by Rails itself.

Regarding dropping support for old Rails versions, I believe it's also fine since Rails [no longer supports any of those](https://guides.rubyonrails.org/maintenance_policy.html).

So my thinking is that it's fine to do this and ship it with the next minor release, but if maintainers prefer major version for this, it's also fine.

I'd very much like the CI to be back to green to avoid further accidental regressions like https://github.com/rails/spring/pull/621.

Thoughts?

deivid-rodriguez:
Hello again!

It's been a couple of months already, so I'd like to offer any help needed to get things moving forward.

kaspth:
Awesome, thank you!

PR #628, opened 2020-10-07
Sequel support

janko:
Like Active Record, Sequel also requires the database connection to be closed before forking. To make Sequel integrate better with Rails out-of-the-box, this PR adds Sequel support to Spring. We don't need to do anything in `#connect_database`, as Sequel will automatically establish a new connection on first query (just like Active Record).

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -291,6 +291,7 @@ module Spring
 
     def disconnect_database
       ActiveRecord::Base.remove_connection if active_record_configured?
+      Sequel::DATABASES.each(&:disconnect) if sequel_configured?
     end
 
     def connect_database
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thank you for the pull request. I support making possible to support Sequel but I don't want to add references to it inside the library. I think it is better to add a new kind of callback, like `on_database_disconnection` and allow apps to write what needs to run in those blocks. Can you change the PR to do that?

janko:
Thanks for the reply. Actually, in the end I've found that using `Spring.after_fork` works just fine for our purposes, and the fact that the `rom-rails` gem does it as well reassures me that it's enough. So, it doesn't seem like there is a need for a separate hook in the end.

PR #629, opened 2020-10-12
Revert "Start server process in dir where cmd was called"

deivid-rodriguez:
This reverts commit 20d3a5f03e0e758a887cfa95d63c0af185d79341.

As discussed with Adrianna, it was never expected that these changes broke existing specs, but we recently found out that they did. Until an alternative solution that doesn't break exisiting specs is found, we agreed on reverting the changes.

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -172,11 +172,6 @@ module Spring
         end
       end
 
-      # Ensure we boot the process in the directory the command was called from,
-      # not from the directory Spring started in
-      original_dir = Dir.pwd
-      Dir.chdir(env['PWD'] || original_dir)
-
       pid = fork {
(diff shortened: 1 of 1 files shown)


adrianna-chang-shopify (approved):
Thank you! ❤️

deivid-rodriguez:
Hello maintainers, thanks for finding time to maintain `spring`.

Is there any chance this PR could get a review? Since it broke existing tests and the original author agreed to revert, it seems quite straightforward :)

I have several open PRs somewhat depending on each other, and I'd like to get things moving so we can improve the default experience when using `spring`.

deivid-rodriguez:
Thanks so much @jeremy! :purple_heart:

Issue #614, opened 2020-05-02
Acceptance test (and Spring.watch) both fail on my macos Catalina


jackkinsella:
Spring's been a real treat over the years, but I noticed it stopped reloading my applications recently. I'm not sure if it is related with my upgrade to Catalina, but the breakage seemed to happen at the same time so I'm suspicious.

Happy to assist in debugging (I'd find it interesting to explore all that socket and threading stuff which I rarely use in my work). Just tell me what I can do.

# Reproduction steps for failing tests in spring repo

Ruby 2.7.1
Spring 2.1.0
MacOS 10.15.3 (19D76) — using an encrypted Apple File System

```
$ rake test:acceptance

1) Failure:
AcceptanceTest#test_changing_the_gems.rb_works [/private/tmp/spring/test/support/acceptance_test.rb:516]:
expected unsuccessful exit status

$ bin/rails runner 'require "sqlite3"'
--- stderr ---
Running via Spring preloader in process 33064
--- log ---
[2020-05-02 17:32:52 +0200] [33056] [watcher:development] check_stale: mtime=1588433569.5504868 < computed=1588433572.379012
[2020-05-02 17:32:52 +0200] [33056] [watcher:development] marked stale, calling listeners: listeners=[#<Proc:0x00007f8785a12f48@/private/tmp/spring/test/apps/gems/2.6.5/gems/spring-2.1.0/lib/spring/application.rb:70>]
[2020-05-02 17:32:52 +0200] [33056] [application:development] running -> watcher_stale
[2020-05-02 17:32:52 +0200] [33056] [application:development] watcher_stale -> exiting
[2020-05-02 17:32:52 +0200] [33055] [application_manager:development] child 33056 shutdown
[2020-05-02 17:32:52 +0200] [33062] [application:development] preloading app
[2020-05-02 17:32:52 +0200] [33062] [watcher:development] start: poller=nil
[2020-05-02 17:32:52 +0200] [33055] [server] accepted client
[2020-05-02 17:32:52 +0200] [33055] [server] running command rails_runner
[2020-05-02 17:32:52 +0200] [33063] [client] sending command
[...]

PR #630, opened 2020-10-16
Fix acceptance test for under rails 5.1

Sixeight:
| before | after |
|--|--|
| image | image |

# Why
- Acceptance test is failed with under Rails 5.1.0 because the Gemfile has [multiple `gem 'sqlite3'` line](https://github.com/rails/spring/blob/a85d32ef3726931b2b81abbb10a6e1c2964e7e32/test/support/application_generator.rb#L61-L70).
- But the test [replace only one line of `gem 'sqlite3'`](https://github.com/rails/spring/blob/a85d32ef3726931b2b81abbb10a6e1c2964e7e32/test/support/acceptance_test.rb#L144) for `gems.rb` changing test.

# What
- Ensure 1 line of `gem 'sqlite3'` . ( [commit](https://github.com/rails/spring/pull/630/commits/fdf0f8b21552f19ded444123a174920e93bad822) )
- Use `gsub` to comment out `gem sqlite3'` .

# Other fails
- <del>`LoadError: cannot load such file -- bump/tasks` </del>
  - Meregd 🎉 fix at https://github.com/rails/spring/pull/629 
- not acceptance test
  - `sprockets requires Ruby version >= 2.5.0. The current ruby version is 2.4.6.354.`
    - Officially support Ruby 2.4.x ?
  - `ArgumentError: wrong number of arguments (given 2, expected 1)` ([here](https://github.com/rails/spring/blob/d7b21ff9afcaf1f5904b7a2286e2a8201728cb54/test/support/watcher_test.rb#L36))
    - [Keyword arguments are now separated from positional arguments](https://bugs.ruby-lang.org/issues/14183) on Ruby 3
    - `FileUtils.touch` cannot accept Hash as optional argument.

# Related
- If remove support for old Rails versions, this PR should be ignored.
  - https://github.com/rails/spring/pull/617

Fixes https://github.com/rails/spring/issues/614

### DIFF
--- a/test/support/acceptance_test.rb
+++ b/test/support/acceptance_test.rb
@@ -510,7 +510,7 @@ module Spring
 
           assert_success %(bin/rails runner 'require "sqlite3"')
 
-          File.write(app.gems_rb, app.gems_rb.read.sub(%{gem 'sqlite3'}, %{# gem 'sqlite3'}))
+          File.write(app.gems_rb, app.gems_rb.read.gsub(%{gem 'sqlite3'}, %{# gem 'sqlite3'}))
           app.await_reload
 
           assert_failure %(bin/rails runner 'require "sqlite3"'), stderr: "sqlite3"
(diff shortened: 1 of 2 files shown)


Sixeight:
merged 🎉 
https://github.com/rails/spring/pull/617

PR #627, opened 2020-10-06
Ensure bundler is required when used

elia:
`Spring.gemfile` is using bundler, the require is needed in certain cases in which Bundler is not already loaded.

Specifically when spring is loaded via `Spring::ApplicationManager` the `#start_child` method wraps the call inside `Bundler.with_original_env` restoring a `RUBYOPT` env variable without the bundler require. That, in turn, causes errors like:

    /Users/elia/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/spring-2.1.1/lib/spring/configuration.rb:10:in `gemfile': uninitialized constant #<Class:Spring>::Bundler (NameError)

Fixes #624

PR #588, opened 2019-03-29
Add support for gems.rb and gems.locked, closes #524.

remomueller:
This pull request merges together @mizalewski's and @JuPlutonic's changes on #538 to support `gems.rb` and `gems.locked` files in spring. I've also wrapped the tests to only run on Ruby 2.0.0 or higher. Let me know if anything else is needed, thanks!

PR #631, opened 2020-10-16
Fix does not watch `gems.locked`

Sixeight:
# Why
- Added `gems.rb` supprt at https://github.com/rails/spring/pull/588 .
- But watcher watches wrong file: `gems.rb.lock` .

# What
- Ensure watch `gems.locked` when using `gems.rb` .

# Related
- https://github.com/rails/spring/pull/627
  - If forked process can depend bundler , Can I use `Bundler::SharedHelpers`'s [methods](https://github.com/rubygems/rubygems/blob/5bb55832ccaa91c229564142c906f959db4dab4c/bundler/lib/bundler/shared_helpers.rb#L13-L32) to shrink loaded file size.
  - If cannot, Should we implement finding _gemfile_ logic?

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -117,7 +117,7 @@ module Spring
       raise e unless initialized?
     ensure
       watcher.add loaded_application_features
-      watcher.add Spring.gemfile, "#{Spring.gemfile}.lock"
+      watcher.add Spring.gemfile, Spring.gemfile_lock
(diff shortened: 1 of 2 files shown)


JuPlutonic:
@rafaelfranca this PR is important for Spring, because provided here method Spring.gemfile_lock will be useful around all the code of Spring project.

PR #640, opened 2021-01-10
Add `rails db:migrate` to Spring-enhanced commands

f1sherman:
I noticed recently that running `bin/rails db:migrate` on our app is much slower than `bin/rake db:migrate` and tracked that down to `db:migrate` not being a Spring-enhanced command. Given that many Rails developers run this command quite often, possibly more often than some of the other Spring-enhanced commands, it seemed to make sense to add it.

I wasn't sure if I should add all of the `db:migrate*` tasks (e.g. `db:migrate:redo`). Since `db:migrate` probably gets run much more often that those others in many use cases this seems to satisfy the 80/20 rule. I'm happy to add those other commands if that's desired.

I didn't see any test coverage for similar commands, let me know if I should add anything.

Thank you!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Next Release
 
+## 2.1.2
+
+* Add `rails db:migrate` to Spring-enhanced commands
+
 ## 2.1.1
 
 * Avoid -I rubylibdir with default-gem bundler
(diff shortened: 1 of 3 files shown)


kaspth:
Nice! Since we'd also want this for `db:rollback` and `db:migrate:redo`, is it possible to teach Spring to look for `bin/rails` where it once looked for `bin/rake` (keeping the `bin/rake` lookup)? Because it must've had that, and that way we'd get it automatically for all the past enriched commands.

f1sherman:
@kaspth great question! I'm sorry but I think I'd need a couple more hours diving in to the source to get a feel for how I might make that change, and I'm not able to commit to that at this time.

Issue #649, opened 2021-08-22
Spring uses AS::Dependencies.mechanism=


fxn:
The accessor `ActiveSupport::Dependencies.mechanism` is private and deleted in Rails 7.

Rails has only one way to enable or disable reloading: `config.cache_classes`. Spring should document it requires that to be false because it reloads, and verify the flag in an initializer for robustness.

byroot:
Ah, I just noticed the same thing today when upgrading Rails and worked on a fix: https://github.com/rails/spring/pull/651

fxn:
@byroot I see there is an additional occurrence in https://github.com/rails/spring/blob/53d9e17d7f125969f21babefd513f502c8c04df5/lib/spring/application.rb#L202.

I don't discard coming later and replacing all this with documentation. Spring has no business manipulating the configuration of the user at its will. That configuration has to be there before we bootstrap, and nobody is telling you what you can do after it.

byroot:
Ah good catch. Sorry I didn't do a great job here, and just directly fixed the error I was seeing.

It's also unfortunate that the true acceptance tests don't run against rails edge, which would likely have caught this.

I agree spring probably shouldn't deal with that. But maybe it could make sense to do a new major version that requires Rails 7? Since it could assume Zeitwerk it might simplify thing. Just a thought.

fxn:
> I agree spring probably shouldn't deal with that. But maybe it could make sense to do a new major version that requires Rails 7? Since it could assume Zeitwerk it might simplify thing. Just a thought.

It's not really about Rails 7. In my view, it should just not do anything of this and be compatible with all versions. You document `config.cache_classes` must be `false` because Spring reloads. If you want to reload, that is `config.cache_classes = false` in Rails.

Then, the initializer aborts if the setting is not `false`.

dhh:
Seems like this is still a problem. @byroot were you planning to have a look at this?

byroot:
If I find the time yes, but it might take a while.

fxn:
I'll do it!

PR #651, opened 2021-08-24
Fix Rails 7.0 compatibility

casperisfine:
Fix: #649 

```
~/gems/spring-2.1.1/lib/spring/application.rb:103:in `block in preload':
undefined method `mechanism=' for ActiveSupport::Dependencies:Module (NoMethodError)
```


fxn:
That is one possible way to do it too, since by now Spring seemed to be working just fine in both modes.

Issue #420, opened 2015-08-13
Need a way to restart server on important environment variable changes


joshuaflanagan:
A lot of people have been bitten by the fact that spring doesn't reload application config (database.yml, config/initializers) with new environment variables. `DATABASE_URL` is a perfect example. There is a PR that attempts to solve part of the problem (https://github.com/rails/spring/pull/267), but even the comments there indicate there is still room for confusing behavior.

We need a way to specify a set of environment variables that should trigger a full reload. By default, it may just include `DATABASE_URL`, but you should be able to add your own environment variables to that list, in case you use them in initializers.

I tried diving into the code to create a PR, but don't have a strong enough understanding of the different environments in play. I wrote a failing test that demonstrates the problem.

``` ruby
test "changing specific environment variables should reload all configuration" do
  File.write(app.path('config/initializers/set_foo.rb'), <<-CONFIG)
     Rails.application.config.foo = ENV['FOO']
  CONFIG

  app.env['FOO'] = 'first'
  assert_success "bin/rails runner 'p Rails.application.config.foo'", stdout: "first"

  app.env['FOO'] = 'second'
  assert_success "bin/rails runner 'p Rails.application.config.foo'", stdout: "second"
end
```

As an example, I can get this to pass by making the following change:

``` ruby
# lib/spring/client/run.rb
def call
  if true # check if important environment variable changed
    stop_server
    cold_run
  elsif env.server_running?
    warm_run
  else
    cold_run
  end
rescue Errno::ECONNRESET
  exit 1
ensure
  server.close if @server
end
```

Obviously, that isn't what we want, since it reboots the server _every_ time. But if we could detect the ENV change at that point, we could conditionally restart everything.

grosser:
are there env vars that change and should not trigger a reload ?

joshuaflanagan:
I don't know. I originally thought it should reload on any ENV change, but wasn't sure what kind of unintended consequences that might have. The full reload is pretty drastic (kills all spring perf benefits), so I figured we should be conservative with it and only use it for values we know we care about.

But I think not-respecting environment variable changes is a major violation of "principle of least astonishment" and people will collectively continue to lose hours of productivity until it is addressed.

grosser:
@jonleighton @rafaelfranca thoughts on this env whitelisting vs blacklisting <-> restart ?

grosser:
[...]

PR #654, opened 2021-09-28
Allow to override environment on spring client startup

### LINKED ISSUE #653: Environment variables changed by rails cannot be reloaded
The following piece of code handle environment variable reloads in Spring

https://github.com/rails/spring/blob/770b2b15b2ecb04fc0c3bda106a89d4a111a3a6c/lib/spring/application.rb#L163-L167

What it does is:

- iterate over environment variables from when spring was started and remove those that rails did not change
- iterate over client environment variables and if those are not present in the environment already, replace them

My problem is that I have a Rails application where there is an environment variable that triggers if the app is in master mode or slave mode. I use ENV! to manage environment variable, and if a variable is not set, ENV! sets a default value. The following sequence of events cause an issue:

- spring server starts without the MASTER environment variable
- app sets MASTER=false which is its default value
- some time later...
- I start the app again with MASTER=true on the command-line
- Spring server receives a client request with MASTER=true
- the process above is followed:
    - it considers MASTER to be unset when the spring server started but it has now MASTER=false. It received a client request with MASTER=true
    - it **does not** removes MASTER from the environment because it sees the variable changed (fron no value to false)
    - it **does not** replace MASTER=false with MASTER=true because it already has a value.
- the app is started in slave mode despite having specified MASTER=true on the command-line.

I don't know the rationale behind this code, but in some cases it is completely wrong.

I believe this might be a reason behing the https://github.com/rails/spring/issues/420 issue

mildred:
Spring.env_override can be set to
- false (default) for current behaviour
- true to always override client provided environment variables in the
  application
- an array listing the environment variables to override in the
  application

The after_fork callback can take an ptional argument that contains the
Spring provided environment to help the application detect when there is
a mismatch between application environment and provided environment.

Fixes #653 and I believe will help with #420 (although a server restart is not performed, only environment overrides)

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -187,13 +187,27 @@ module Spring
         # Delete all env vars which are unchanged from before Spring started
         original_env.each { |k, v| ENV.delete k if ENV[k] == v }
 
+        # if env_override is an array, delete the variables listed
+        if Spring.env_override.is_a? Array
+          Spring.env_override.each { |k| ENV.delete k }
+        end
+
         # Load in the current env vars, except those which *were* changed when Spring started
-        env.each { |k, v| ENV[k] ||= v }
+        # if env_override is true, load in *all* the current vars
+        env.each do |k, v|
+          if Spring.env_override == true
+            ENV[k] = v
+          else
+            ENV[k] ||= v
+          end
+        end
 
         connect_database
         srand
 
-        invoke_after_fork_callbacks
+        invoke_after_fork_callbacks(OpenStruct.new(
+          env: env,
+          original_env: original_env))
         shush_backtraces
 
         command.call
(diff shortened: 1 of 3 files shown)


lib/spring/application.rb
```diff
@@ -187,13 +187,27 @@ def serve(client)
...
+          else
+            ENV[k] ||= v
+          end
+        end
 
         connect_database
         srand
 
-        invoke_after_fork_callbacks
+        invoke_after_fork_callbacks(OpenStruct.new(
```
rafaelfranca, line 208: Can we create a class for this object here? It would allow us to document it for example.

rafaelfranca:
Closed by https://github.com/rails/spring/pull/704

PR #656, opened 2021-11-24
Drop dependency on StringIO

casperisfine:
StringIO was extracted as a gem, so spring can no
longer use it for the same reason it can't use the `json`
gem.

Otherwise you'll see warnings like:

```
...stringio.bundle: warning: already initialized constant StringIO::VERSION
```

This can be reproduced on Ruby 3.0, by installing `stringio 3.0.1`
and starting a Rails app with spring.

cc @rafaelfranca

### DIFF
--- a/lib/spring/json.rb
+++ b/lib/spring/json.rb
@@ -46,8 +46,6 @@ end
 
 # See https://github.com/kr/okjson for updates.
 
-require 'stringio'
-
 # Some parts adapted from
 # https://golang.org/src/pkg/json/decode.go and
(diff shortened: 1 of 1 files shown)


lib/spring/json.rb
```diff
@@ -46,8 +46,6 @@ def self.dump(data)
 
 # See https://github.com/kr/okjson for updates.
```
casperisfine, line 47: The upstream repo is archived so I think it's no big deal to modify it.

Also I wonder why `Marshal` wasn't used in the first place.

PR #659, opened 2021-12-02
Stop depending on Set

casperisfine:
It is now a gem, and since spring load before bundler if you have multiple `set` versions installed, you might load the wrong one causing already defined constant warnings.

Initially wrongly reported as a bundler issue: https://github.com/rubygems/rubygems/issues/5127

Reported to be my @Bowman-42

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -11,8 +11,8 @@ module Spring
       @original_env = original_env
       @spring_env   = spring_env
       @mutex        = Mutex.new
-      @waiting      = Set.new
-      @clients      = Set.new
+      @waiting      = {}
+      @clients      = {}
       @preloaded    = false
       @state        = :initialized
       @interrupt    = IO.pipe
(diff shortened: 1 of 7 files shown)


lib/spring/application.rb
```diff
@@ -181,7 +181,7 @@ def serve(client)
       pid = fork {
         # Make sure to close other clients otherwise their graceful termination
         # will be impossible due to reference from this fork.
-        @clients.select { |c| c != client }.each(&:close)
+        @clients.each_key { |c| c.clonse if c != client }
```
deivid-rodriguez, line 184:
```suggestion
        @clients.each_key { |c| c.close if c != client }
```

casperisfine: Woops, thank you :)

casperisfine:
cc @rafaelfranca, this isn't a big bug as it's still very rare to have `set` installed as a gem, but would be nice to have it released.

PR #665, opened 2021-12-22
Stop requiring Set

nvasilevski:
This is an addition to the https://github.com/rails/spring/pull/659 PR
We stopped using `Set` so it should be unnecessary to require it


byroot:
It's beyond me how I managed to miss theses 🤦

Issue #603, opened 2019-11-06
Different version of standard gems loaded by Spring and Bundler


scottmacpherson:
As Spring loads before Bundler and therefore relies on vanilla RubyGems to load dependencies, Spring can end up loading different versions of standard gems from the main Rails application.

This is noticable with FileUtils: version 1.1.0 shipped with Ruby 2.6.5, but the latest version is 1.3.0:

    $ gem list fileutils
    fileutils (1.3.0, default: 1.1.0)
    $ ruby --version
    ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin18]
    $ spring --version
    Spring version 2.1.0

RubyGems defaults to loading the latest installed version of a required gem so when Spring boots and requires FileUtils, version 1.3.0 is loaded. However once inside the full Rails application and in the hands of Bundler, requiring FileUtils returns version 1.1.0. (Unless declared in Gemfile, Bundler will load the version of a stdlib gem that shipped with Ruby.)

This in turn leads to warnings about constants already being defined, e.g.:

> /Users/scott/.rbenv/versions/2.6.5/lib/ruby/2.6.0/fileutils/version.rb:4: warning: already initialized constant FileUtils::VERSION
> /Users/scott/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/fileutils-1.3.0/lib/fileutils/version.rb:4: warning: previous definition of VERSION was here

This particular issue can be overcome by adding FileUtils to Gemfile, but Spring obviously relies on other gemified stdlib gems which may result in other subtle differences when running Rails with and without Spring.

scottmacpherson:
My mistake. @aar0nr has pointed out in [ruby\fileutils#22](https://github.com/ruby/fileutils/issues/22#issuecomment-566118203) that this is likely caused by Bootsnap ignoring a bundle, and Spring is only caught up because it's the first thing in a new Rails application that loads FileUtils by name.

PR #672, opened 2022-02-06
Lazy-load digest dependency

theodorton:
I've experienced some pain in running spring and bootsnap together. Symptom is a `Gem::LoadError` thrown when there are conflicting versions of digest installed:

> You have already activated digest 3.0.0, but your Gemfile requires digest 3.1.0. Since digest is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports digest as a default gem. (Gem::LoadError)

Since `digest` is now being included as a default gem in Ruby, it's hard to circumvent the issue without locking `digest` to an older version.

The problem seems to disappear when deferring the require call.

Relevant issues to similar problems:
https://github.com/rails/spring/issues/603
https://github.com/rails/spring/pull/616

### DIFF
--- a/lib/spring/env.rb
+++ b/lib/spring/env.rb
@@ -1,5 +1,4 @@
 require "pathname"
-require "digest/md5"
 
 require "spring/version"
 require "spring/configuration"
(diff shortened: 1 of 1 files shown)


Tonkpils:
@rafaelfranca any chance we can backport this to the Spring 3 release? Upgrading to Rails 7 seems to trigger this failure and a few gems like `spring-watcher-listen` have a restriction on spring of `< 4.0`.

yjukaku:
It would be helpful as well for us to have this backported to Spring 3.

PR #674, opened 2022-05-02
Add support for SPRING_QUIET environment variable.

### LINKED ISSUE #467: Spring stderr output causing warnings in other apps
This affects RubyMine currently, but possibly others. It detects the output to stderr and thinks it's a problem with the command it's trying to run. I understand the motivation behind putting it on stderr, so I'm not sure I have a good answer.

Related: I see that I can set `Spring.quiet = true` to silence the output, but the documentation doesn't indicate where to do that. Where should I be putting that?

thewoolleyman:
Adds support for enabling quiet mode via `SPRING_QUIET` environment variable.

This is useful in situations where you don't want to put `Spring.quiet = true` in `~/.spring.rb` or your app's `config/spring.rb`, but want to enable quiet mode on a case-by-case basis.

See more details in the doc updates in this PR.

Closes #467

### DIFF
--- a/lib/spring/configuration.rb
+++ b/lib/spring/configuration.rb
@@ -2,7 +2,8 @@ require "spring/errors"
 
 module Spring
   class << self
-    attr_accessor :application_root, :quiet
+    attr_accessor :application_root
+    attr_writer :quiet
 
     def gemfile
       require "bundler"
@@ -50,6 +51,10 @@ module Spring
 
     def project_root_path
       @project_root_path ||= find_project_root(Pathname.new(File.expand_path(Dir.pwd)))
+    end
+
+    def quiet
+      @quiet ||= ENV.key?('SPRING_QUIET') 
     end
 
     private
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next Release
 
+* Add support for `SPRING_QUIET` environment variable.
+
 ## 4.0.0
 
 * Stop requiring `set` before bundler can select the proper version. This could result in
--- a/README.md
+++ b/README.md
@@ -388,6 +388,12 @@ a command runs:
 Spring.quiet = true
 ```
 
+You can also set the initial state of the `quiet` configuration option to true
+by setting the `SPRING_QUIET` environment variable before executing Spring.
+This is useful if you want to set quiet mode when invoking the Spring executable
+in a subprocess, and cannot or prefer not to set it programmatically
+via the `Spring.quiet` option in `~/.spring.rb` or the app's `config/spring.rb`.
+
 ### Environment variables
 
 The following environment variables are used by Spring:
@@ -412,6 +418,8 @@ The following environment variables are used by Spring:
   the long-running Spring server process. By default, this is related to
   the socket path; if the socket path is `/foo/bar/spring.sock` the
   pidfile will be `/foo/bar/spring.pid`.
+* `SPRING_QUIET` - If set, the initial state of the `Spring.quiet`
+  configuration option will default to `true`.
 * `SPRING_SERVER_COMMAND` - The command to run to start up the Spring
   server when it is not already running. Defaults to `spring _[version]_
   server --background`.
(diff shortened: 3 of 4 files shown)


thewoolleyman:
I just noticed that @bobf already had a PR to implement this in https://github.com/rails/spring/pull/577

However, I'm leaving this one open, because:

1. It has more extensive docs
2. It short-circuits in the config reader method via `||=` instead of `||`, allowing it to be set back to `false` even after spring starts, if this is necessary.

If this PR is merged, #577 can be closed as a duplicate.

thewoolleyman:
@rafaelfranca , could we get please some attention on this one to get it merged? It will close out the following:

- https://github.com/rails/spring/pull/674
- https://github.com/rails/spring/pull/577
- https://github.com/rails/spring/issues/467

Thanks! 🙏

bobf:
Glad to see some movement on this. Please feel free to close my PR once this one is closed, or if this one ends up getting stalled somehow then I'm happy to merge the changes into my PR and make any adjustments needed.

thewoolleyman:
Hi @tenderlove 👋!  any chance you or some other committer could help get some attention to getting this merged? Thanks!

lib/spring/configuration.rb
```diff
@@ -52,6 +53,10 @@ def project_root_path
       @project_root_path ||= find_project_root(Pathname.new(File.expand_path(Dir.pwd)))
     end
 
+    def quiet
+      @quiet ||= ENV.key?('SPRING_QUIET')
```
nvasilevski, line 57: >It short-circuits in the config reader method via ||= instead of ||, allowing it to be set back to false even after spring starts, if this is necessary.

Correct me if I'm wrong, but at first glance it looks like that it won't be possible to set it back to `false` after setting it to `true` once, like:
```ruby
class DummySpring
  class << self
    attr_writer :quiet

    def quiet
      @quiet ||= ENV.key?('SPRING_QUIET') 
    end
  end
end

ENV["SPRING_QUIET"] = "lol"
DummySpring.quiet # => true
DummySpring.quiet = false
DummySpring.quiet # => still true because `ENV.key?('SPRING_QUIET') ` gets checked again as long as `@quiet` is falsey
```

thewoolleyman: @nvasilevski  That's a good point. You could set it back to false via the `attr_writer`, but not via the `ENV` var.

But I guess I assumed that you wouldn't try to use the ENV var to change it while an app is executing - if you want to do that, just use the `attr_writer`.

In other words, the `ENV` var is only needed when you *_don't_* have access to modify the code or config at all, but still want to turn quiet mode on,

This is also explained by the new documentation:

> You can also set the initial state of the `quiet` configuration option to true
by setting the `SPRING_QUIET` environment variable before executing Spring.

Note it says "initial state" and "...before executing Spring".

So, I guess I think this is OK as is, especially given the documentation.

WDYT?

nvasilevski: >You could set it back to false via the attr_writer, but not via the ENV var.

In my example I'm showing that it's not possible to reset it using `writer` after setting it once to `true` because of how `||=` works. If we want to support that, reader will need to be implemented as:
```ruby
def quiet
    return @quiet if defined(@quiet)

    @quiet = ENV.key?('SPRING_QUIET')
end
```

thewoolleyman: > In my example I'm showing that it's not possible to reset it using writer after setting it once to true because of how ||= works.

Ah, gotcha, you're right.

So ideally we can come up with a case where it can be changed from true to false or vice versa, via either the attr_writer or the ENV var.

I'll try to come up with a change that does that and then ping you to review it.

Thanks for your help and attention!

thewoolleyman: Sorry, I got busy/lazy and never returned to finish up this thread.

But I see this got merged now, which is great! 

This thread describes a minor bug which should be fixed eventually, so I'll keep it on my TODO list - near the bottom, which realistically means "probably never", but hope springs eternal ;)

PR #684, opened 2022-10-05
Run without error if spring is in uninstalled group

mkllnk:
We couldn't run the console on our production server since updating our binstubs with #662. Our setup has not been considered in that change but we found an easy change to the spring binstub which I'm sharing here.

We have spring in the development and test groups which means that it's present in the Gemfile but not installed on production. We are also not using the RAILS_ENV variable when logging in as developer and start a console with a parameter:

    ./bin/rails console -e production

In this case, a Gem::LoadError was raised when trying to load spring. So since checking the RAILS_ENV is not a reliable detection of spring being installed, we added a rescue call.

But looking into this issue further, I discovered that we can go through the list of actually requested gems and skip spring if it wasn't requested. This solves our specific case. A rescue statement may still be useful if other people have other reasons for the gem not being installed but I don't know if that's possible. An alternative or additional patch is available if you are interested but it may not be necessary:  https://github.com/rails/spring/compare/main...mkllnk:handle-missing-gem

### DIFF
--- a/test/support/acceptance_test.rb
+++ b/test/support/acceptance_test.rb
@@ -76,9 +76,12 @@ module Spring
 
       def without_gem(name)
         gem_home = app.gem_home.join('gems')
+        spec_home = app.gem_home.join('specifications')
         FileUtils.mv(gem_home.join(name), app.root)
+        FileUtils.mv(spec_home.join("#{name}.gemspec"), app.root)
         yield
       ensure
+        FileUtils.mv(app.root.join("#{name}.gemspec"), spec_home)
         FileUtils.mv(app.root.join(name), gem_home)
       end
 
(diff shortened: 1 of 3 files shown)


test/support/acceptance_test.rb
```diff
@@ -77,9 +77,12 @@ def assert_speedup(ratio = DEFAULT_SPEEDUP)
 
       def without_gem(name)
         gem_home = app.gem_home.join('gems')
+        spec_home = app.gem_home.join('specifications')
         FileUtils.mv(gem_home.join(name), app.root)
+        FileUtils.mv(spec_home.join("#{name}.gemspec"), app.root)
```
mkllnk, line 82: Without moving the spec file, `gem "spring", spring.version` succeeds even though the gem code isn't there. The the next line raises a `LoadError` instead of a `Gem::LoadError`. That doesn't really matter for this solution but was an issue when trying to rescue only the `Gem::LoadError`. Anyway, I thought that this is more realistic for future specs.

PR #686, opened 2022-12-05
Update readme for `enable_reloading` setting

ghiculescu:
ref: https://github.com/rails/rails/pull/46648


janko:
Isn't `config.enable_reloading` scheduled to be added in Rails 7.1? I remember trying it out in Rails 7.0, and `config.enable_reloading = ` didn't have any effect (Rails appears to allow calling `config.anything =`).

ghiculescu:
🤦 sorry, you're right!

PR #692, opened 2023-01-31
Documentation should reflect reality

pboling:
Documentation of created bin stub in README.md is wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,17 +48,10 @@ This generates a `bin/spring` executable, and inserts a small snippet of
 code into relevant existing executables. The snippet looks like this:
 
 ``` ruby
-begin
   load File.expand_path('../spring', __FILE__)
-rescue LoadError => e
-  raise unless e.message.include?('spring')
-end
 ```
 
-On platforms where Spring is installed and supported, this snippet
-hooks Spring into the execution of commands. In other cases, the snippet
-will just be silently ignored, and the lines after it will be executed as
-normal.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,17 +48,10 @@ This generates a `bin/spring` executable, and inserts a small snippet of
...
-rescue LoadError => e
-  raise unless e.message.include?('spring')
-end
 ```
 
-On platforms where Spring is installed and supported, this snippet
-hooks Spring into the execution of commands. In other cases, the snippet
-will just be silently ignored, and the lines after it will be executed as
-normal.
+This will cause the application to fail to load on platforms which do not install the Spring gem, e.g. Production.
```
rafaelfranca, line 54: Yeah, this seems incorrect. Did you try it? Spring isn't loaded at all in production https://github.com/rails/spring/pull/662/files#diff-75bd83565389d1cdfbd3d56c3414272fadd7aba5e9d0a892059bce5dcaea0fabR28

pboling: I think I get it now.  The documentation needs to be updated to reflect that^ and also to mirror the contents of the udpated bin stub...  So this PR is not quite right, but there is an issue in the current readme in latest `main` branch, where it doesn't match what is being created as of #662.

pboling:
Closing because this change is incorrect, but the README was *also* incorrect at this time. I'm not sure if the documentation has been, or will be, fixed to reflect reality.

PR #701, opened 2023-08-16
DEPRECATION WARNING: Using a :default format for Time#to_s is depreca…

moskvin:
…ted. Please use Time#to_fs instead.

### DIFF
--- a/lib/spring/env.rb
+++ b/lib/spring/env.rb
@@ -79,8 +79,13 @@ module Spring
       end
     end
 
+    def time_now
+      time = Time.now
+      time.respond_to?(:to_fs) ? time.to_fs : time.to_s
+    end
+
     def log(message)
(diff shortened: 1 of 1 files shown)


leonid-shevtsov:
IMO it makes more sense to do `Time.now.iso8601`; that's what the default format for `to_s` was, and it works well for log timestamps. And is more explicit.

matthewd:
It's not this library's job to protect itself against monkey-patched core methods.

Correspondingly, if someone wants to monkey-patch `to_s` (the effective result of configuring a `:default` format), they should do that directly -- at which point this call will be duly affected just like any other.

leonid-shevtsov:
>It's not this library's job to protect itself against monkey-patched core methods.

This is a Rails 7 issue, how is it not Spring's job to be compatible with Rails?

PR #703, opened 2023-11-14
Drop dependency on mutex_m

byroot:
It doesn't really save much effort, and it's being extracted as a gem which cause issues for projects like spring that often are loaded before bundler.

Reported by @tisba

### DIFF
--- a/lib/spring/watcher/abstract.rb
+++ b/lib/spring/watcher/abstract.rb
@@ -1,5 +1,4 @@
 require "pathname"
-require "mutex_m"
 
 module Spring
   module Watcher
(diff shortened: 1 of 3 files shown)


fschwahn:
@byroot This breaks eg. `spring-watcher-listen` (see https://github.com/rails/spring-watcher-listen/issues/36), as it depends on the synchronize method. Possibly there are other dependencies affected by this.

Maybe instead of patching all dependencies which depend on this synchronize method could be added back to `Abstract`?

```
def synchronize(&block)
  @mutex.synchronize(&block(
end
```

byroot:
Thanks for the ping.

fschwahn:
Thank you for the quick turnaround 👍

PR #705, opened 2023-11-22
Add Spring.spawn_on_env

gmcgibbon:
Similar to https://github.com/rails/spring/pull/704, but spawns a new app instead of restarting the app process.

Currently, the only way Spring can create multiple applications for a given environment variable value is by changing the `SPRING_APPLICATION_ID`. This has the tradeoff of not being visible in `bin/spring status` calls, and being treated as a separate application, when it actually isn't. You might want to run an application with and without certain features enabled in the same rails env, and with this patch you can do that and have Spring treat these as the same app.

Configuration must be set in `config/spring_client.rb`:

```rb
Spring.spawn_on_env << "SOME_ENV"
```

```sh
> SOME_ENV=1 bin/rails runner "" # start server
> SOME_ENV=2 bin/rails runner "" # starts another server
> bin/spring status
Spring is running:

<pid> spring server | app | started 28 secs ago
<pid> spring app    | app | started 28 secs ago | development mode
<pid> spring app    | app | started 28 secs ago | development mode | RAILS_CONTEXT=essentials
```

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -28,6 +28,11 @@ module Spring
       @interrupt.last.write "."
     end
 
+    def spawn_env
+      env = JSON.load(ENV["SPRING_SPAWN_ENV"].dup).map { |key, value| "#{key}=#{value}" }
+      env.join(", ") if env.any?
+    end
+
     def app_env
       ENV['RAILS_ENV']
     end
--- a/lib/spring/application/boot.rb
+++ b/lib/spring/application/boot.rb
@@ -11,9 +11,15 @@ app = Spring::Application.new(
 
 Signal.trap("TERM") { app.terminate }
 
-Spring::ProcessTitleUpdater.run { |distance|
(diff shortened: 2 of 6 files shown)


byroot:
> Currently, it isn't possible for Spring to create multiple applications for a given environment variable value.

I believe it's possible via `ENV["SPRING_APPLICATION_ID"]`

gmcgibbon:
Ok, yes it is possible via `ENV["SPRING_APPLICATION_ID"]`, but it doesn't list all the applications in `bin/spring status`, and prefixing environment variables and manually crafting the spring app id with them isn't straightforward IMO. They are all the same application, so they should all be visible to spring calls by default.

gmcgibbon:
@byroot if you have time, do you have any more feedback?

byroot (approved):

gmcgibbon:
Hm, looks like CI is failing for a specific version for some dependency reasons. I'll look into that separately, I don't think it is related to this change.

Issue #663, opened 2021-12-13
Processes just hang with Spring 4.0


coding-red-panda:
Hello,

With the release of Spring 4.0, rake tasks and processes just hang, both on local systems as well as the CI Pipeline.
For example:

```bash
bundle exec rails db:drop db:create db:schema:load
```

The above will just run indefinitely, with no output or any error messages.
Does not respond to CTRL+C interrupts and requires the usage of `kill` to actually terminate the Ruby processes.

However when I switch back to Spring 3.1.1 everything worked:

```bash
olivar@Arnes-MacBook-Pro  ~/RubymineProjects/deployment_history   master  bundle exec rake db:drop
Dropped database 'deployment_history_development'
olivar@Arnes-MacBook-Pro  ~/RubymineProjects/deployment_history   master  bundle exec rake db:create db:schema:load
Created database 'deployment_history_development'
```

This is on a Rails 6.1 Project with Ruby 2.7.3

juhguu:
I have the same issue. 
I'm using ubuntu 20.04, Rails 6.1.4.1 and ruby 3.0.1.
The only command that works for me is `rails server`.
When I remove the Spring of my project or switch back to Spring 3.x everything worked.

kwstannard:
Same but on Rails 5.2

gabrielboeker:
has been anything done since then? i was just tinkering around for 45 minutes to get my app running again since the rails commands were not executed anymore. downgraded spring to 2.1.1 to make it work again. im on Rails 6.1.4. and Ruby 2.7.4. Spring was on 4.0.0 which was broken

technicalpickles:
I've seen similar issues before. `spring stop` can help, but it often occurs again. Just now when I ran into it, I poked around with `ps -ef | grep spring`, and had several spring processes from the past few days. I've killed those, and it seems to be looking better.

Anyone have ideas on what type of data to collect the next time I see it?

timdiggins:
I've had this problem and for me the problem was I was using Bundler v1 which doesn't provide `Bundler.with_unbundled_env` which Spring 4 uses. Upgrading to Bundler v2 for this project has solved this (https://bundler.io/guides/bundler_2_upgrade.html)

FYI to debug this and find the problem I did the following:

* Start spring as a foreground process (so you can see the issue) in one terminal session using `spring server`. 
* Then start a rails process (like `rails c`) in another terminal session. 
* The "spring server" quits with a traceback, showing the problem.

[...]

PR #708, opened 2023-12-06
Add timeout when connecting to server

technicalpickles:
Over on https://github.com/rails/spring/issues/663 we've seen the spring server get into a funky state, where the `server.gets` just hangs indefinitely.

I found other places in client.rb that checked that there was something to read before reading, and made sure `server.gets` returns something. So, applied that to `verify_server_version` 😁

### DIFF
--- a/lib/spring/client/run.rb
+++ b/lib/spring/client/run.rb
@@ -114,6 +114,10 @@ module Spring
       end
 
       def verify_server_version
+        unless IO.select([server], [], [], CONNECT_TIMEOUT)
+          raise "Error connecting to Spring server"
+        end
+
         server_version = server.gets.chomp
         if server_version != env.version
           $stderr.puts "There is a version mismatch between the Spring client " \
--- a/lib/spring/client/run.rb
+++ b/lib/spring/client/run.rb
@@ -118,7 +118,12 @@ module Spring
           raise "Error connecting to Spring server"
         end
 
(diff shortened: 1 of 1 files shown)


technicalpickles:
I tested this two ways:

1. `bin/spring start` in one console. background it, and run `bin/rails console` (or any other binstub)
2. `socat unix-listen:tmp/spring.sock /dev/null` in one console. set `SPRING_PIDFILE=tmp/spring.pid SPRING_SOCKET=tmp/spring.sock`, and add the socat pid to tmp/spring.sock

In both, the binstub raises an error as expected.

casperisfine:
cc @byroot

lib/spring/client/run.rb
```diff
@@ -114,7 +114,16 @@ def stop_server
       end
 
       def verify_server_version
-        server_version = server.gets.chomp
+        unless IO.select([server], [], [], CONNECT_TIMEOUT)
```
byroot, line 117: So note that this only tell us there's something to read, it doesn't guarantee it's a whole line, so we can still potentially get stuck in `server.gets`.

However it should catch a large portion of such issues, so I think it's a good improvement.

technicalpickles:
Thank you! I'll be testing this on our app, and I'm hoping it will help surface when https://github.com/rails/spring/issues/663 happens

PR #712, opened 2024-04-16
Allow to configure boot and connect timeout

byroot:
The hardcoded values might not be enough for all applications.

### DIFF
--- a/lib/spring/client/run.rb
+++ b/lib/spring/client/run.rb
@@ -6,8 +6,6 @@ module Spring
   module Client
     class Run < Command
       FORWARDED_SIGNALS = %w(INT QUIT USR1 USR2 INFO WINCH) & Signal.list.keys
-      CONNECT_TIMEOUT   = 1
-      BOOT_TIMEOUT      = 20
 
       attr_reader :server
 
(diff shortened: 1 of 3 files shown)


lib/spring/client/run.rb
```diff
@@ -74,7 +72,7 @@ def boot_server
         env.socket_path.unlink if env.socket_path.exist?
 
         pid     = Process.spawn(server_process_env, env.server_command, out: File::NULL)
-        timeout = Time.now + BOOT_TIMEOUT
+        timeout = Time.now + String.boot_timeout
```
jenshenny, line 75:
```suggestion
        timeout = Time.now + Spring.boot_timeout
```

lib/spring/configuration.rb
```diff
@@ -5,6 +5,12 @@ class << self
     attr_accessor :application_root
     attr_writer :quiet
 
+    attr_accessor :connect_timeout
+    @connect_timeout = 5
+
+    attr_accessor :boot_timeout
+    @boot_timeout = 20
```
jenshenny, line 12: These instance variables aren't being correctly set currently (the accessors return nil)

```
`+': can't convert NilClass into an exact number (TypeError)
```
It works if you place them outside of the class methods

```
module Spring
  @connect_timeout = 5
  @boot_timeout = 20

  class << self
```

jenshenny (commented):
Thanks! Other than the comment I put with the accessors, LGTM

lib/spring/configuration.rb
byroot: Yeah, sorry, just did a quick patch without testing before leaving my desk. I think it should be good now.

Issue #716, opened 2024-04-26
Spring 4.2.0 crashing when running rake


fschwahn:
This happened to me when running `rails db:migrate`.

I have never seen this before, and I have been using spring daily for years. Maybe it is connected to the changes in 4.2.0 (namely #708).

I'm running spring server as a separate process, so I have logs / stacktrace to share:
```
15:50:59 spring.1 | [2024-04-26 15:50:59 +0200] [4120] [server] accepted client
15:50:59 spring.1 | [2024-04-26 15:50:59 +0200] [4120] [server] running command rake
15:50:59 spring.1 | [2024-04-26 15:50:59 +0200] [4120] [server] shutting down
15:50:59 spring.1 | [2024-04-26 15:50:59 +0200] [4120] [application_manager:development] stopping
15:50:59 spring.1 | [2024-04-26 15:50:59 +0200] [4120] [application_manager:test] stopping
15:50:59 spring.1 | [2024-04-26 15:50:59 +0200] [12426] [application:development] forced exit
15:50:59 spring.1 | /Users/fabian/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/spring-4.2.0/lib/spring/application_manager.rb:63:in `send_io': Invalid argument - sendmsg(2) (Errno::EINVAL)
15:50:59 spring.1 | 	from /Users/fabian/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/spring-4.2.0/lib/spring/application_manager.rb:63:in `block in run'
15:50:59 spring.1 | 	from /Users/fabian/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/spring-4.2.0/lib/spring/application_manager.rb:44:in `block in with_child'
15:50:59 spring.1 | 	from /Users/fabian/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/spring-4.2.0/lib/spring/application_manager.rb:22:in `synchronize'
15:50:59 spring.1 | 	from /Users/fabian/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/spring-4.2.0/lib/spring/application_manager.rb:41:in `with_child'
15:50:59 spring.1 | 	from /Users/fabian/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/spring-4.2.0/lib/spring/application_manager.rb:62:in `run'
[...]

PR #727, opened 2024-09-12
Prevent server crash by restarting child

prognostikos:
Before this change, if Errno::EINVAL is thrown when sending IO to the child the server would crash. Now we catch the exception and, assuming the child has problems, we start a new child.

### DIFF
--- a/lib/spring/application_manager.rb
+++ b/lib/spring/application_manager.rb
@@ -42,7 +42,7 @@ module Spring
         if alive?
           begin
             yield
-          rescue Errno::ECONNRESET, Errno::EPIPE
+          rescue Errno::ECONNRESET, Errno::EPIPE, Errno::EINVAL
             # The child has died but has not been collected by the wait thread yet,
             # so start a new child and try again.
             log "child dead; starting"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next Release
 
+* Fixed a bug that would crash the server if sending IO to the child failed
+
 ## 4.2.1
 
 * Added `Spring.connect_timeout` and `Spring.boot_timeout` to allow to increase timeout for larger apps.


prognostikos:
This is a solution, or maybe better characterized as a work-around, for #716 and maybe #663 and #396

Before this change, in a Rails `7.0.8.4` app using `ruby 3.3.4 (2024-07-09 revision be1089c8ec) [arm64-darwin23]` I could trigger the error as follows:

1. start spring using `spring server` in one terminal
2. start `rails c` in a second terminal
3. run `rails db:migrate:status` in a third terminal - it succeeds
4. run `touch config/initializers/content_security_policy.rb` in the third terminal
5. run `rails db:migrate:status` in the third terminal - it hangs with no output and the console running in the second terminal is dead.
6. in the first terminal, spring dies with ```lib/spring/application_manager.rb:64:in `send_io': Invalid argument - sendmsg(2) (Errno::EINVAL)```

I have put the spring server logs for both the error and fixed versions here: https://gist.github.com/prognostikos/3827d79e8e58e220cf32ec3e35684552

I would love to get this into a new Spring release, what needs to be done for that to happen? I can see there are some issues with CI, and running the acceptance tests locally with RAILS_VERSION="edge" also seems to hang so there is probably a bit of work to do on the test setup.

lib/spring/application_manager.rb
```diff
@@ -42,7 +42,7 @@ def with_child
         if alive?
           begin
             yield
-          rescue Errno::ECONNRESET, Errno::EPIPE
+          rescue Errno::ECONNRESET, Errno::EPIPE, Errno::EINVAL
```
prognostikos, line 45: Should we do some kind of cleanup here in case the child is hanging around as a zombie process?

fschwahn:
This seems to fix the problems I encountered in #716 👍

fmichaut-diff:
Could this get attention from the Maintainers ?  It's a very simple patch that solves a very annoying problem

Issue #735, opened 2025-03-19
Spring-enabled binstub fails to start if ENV contains non-ASCII value


mattbrictson:
This bug was introduced in spring 3.4.0.

If the shell environment contains a non-ASCII value, e.g.

```
export HOMEBREW_INSTALL_BADGE=🍵
```

Then running a spring-enabled binstub like `bin/rspec` crashes with an `Encoding::CompatibilityError` error:

```
$ bin/rspec spec/example_spec.rb
gems/spring-4.3.0/lib/spring/json.rb:382:in 'String#[]=': incompatible character encodings: UTF-8 and BINARY (ASCII-8BIT) (Encoding::CompatibilityError)
  from gems/spring-4.3.0/lib/spring/json.rb:382:in 'Spring::OkJson#unquote'
  from gems/spring-4.3.0/lib/spring/json.rb:306:in 'Spring::OkJson#strtok'
  from gems/spring-4.3.0/lib/spring/json.rb:272:in 'Spring::OkJson#tok'
  from gems/spring-4.3.0/lib/spring/json.rb:235:in 'Spring::OkJson#lex'
  from gems/spring-4.3.0/lib/spring/json.rb:66:in 'Spring::OkJson#decode'
  from gems/spring-4.3.0/lib/spring/json.rb:16:in 'Spring::JSON.load'
  from gems/spring-4.3.0/lib/spring/application.rb:185:in 'Spring::Application#serve'
  from gems/spring-4.3.0/lib/spring/application.rb:158:in 'block in Spring::Application#run'
  from <internal:kernel>:168:in 'Kernel#loop'
  from gems/spring-4.3.0/lib/spring/application.rb:152:in 'Spring::Application#run'
  from gems/spring-4.3.0/lib/spring/application/boot.rb:25:in '<top (required)>'
  from <internal:rubygems/core_ext/kernel_require.rb>:136:in 'Kernel#require'
  from <internal:rubygems/core_ext/kernel_require.rb>:136:in 'Kernel#require'
  from -e:1:in '<main>'
```

Downgrading to spring 4.2.1 fixes the problem.

adigitalnative:
I am seeing this as well. My traceback:

```
/Users/jacquelinechenault/.rvm/gems/ruby-3.3.1/gems/spring-4.3.0/lib/spring/json.rb:382:in `[]=': incompatible character encodings: UTF-8 and ASCII-8BIT (Encoding::CompatibilityError)
[...]

PR #736, opened 2025-03-24
Fixed to do force_encoding when decoding JSON.

masato-hi:
### What I did.

Encode variables a and q to UTF-8 during JSON decoding.
Ruby's ENV variable is ASCII-8BIT, so force_encoding is required when accepting values passed via ENV.

### What I didn't do.

Unicode Escape when encoding.
I didn't do this because the decoding process only supports Unicode up to 4 bytes.

### Related.

https://github.com/rails/spring/issues/735

https://github.com/rails/spring/commit/1ed51d04754d782d70398c75000d2d23877326e9#diff-c105c6f17eabaabdafb33fb6b9dccbaaf06e70daba153d42e3e72741a7ff6eccL16


rafaelfranca:
Fixed by #738

PR #740, opened 2025-05-02
Shush the `backtrace_location`, too

### LINKED ISSUE #739: `#shush_backtraces` doesn't clean the `backtrace_locations`
The `Spring::Application#shush_backtraces` method strips out Spring stackframes from the exception's `backtrace`, but not its `backtrace_locations`:

https://github.com/rails/spring/blob/046d50a334bb70e7d5f68c418bf137de4653be9b/lib/spring/application.rb#L322

This causes at least 2 issues:

1. The `backtrace` and `backtrace_locations` become out of sync, and only when running in Spring.
2. It breaks the downstream behaviour of `ActiveSupport::ErrorReporter`, whose `#report` method will clean itself off the top of the stack. Since the top of the stack will be these 2 Spring frames, its logic won't remove any of the frames.

amomchilov:
Fixes #739

Didn't change any tests, because I couldn't find any tests that covered this behaviour.


lib/spring/application.rb
```diff
@@ -319,7 +319,10 @@ def shush_backtraces
             ensure
               if $!
                 lib = File.expand_path("..", __FILE__)
-                $!.backtrace.reject! { |line| line.start_with?(lib) } unless $!.backtrace.frozen?
+                unless $!.backtrace.frozen?
+                  $!.backtrace.reject! { |line| line.start_with?(lib) }
+                  $!.backtrace_locations.reject! { |line| line.path.start_with?(lib) }
+                end
               end
```
amomchilov, line 326: How do we feel about giving `$!` a more reasonable name?

```suggestion
              error = $!
              if error
                lib = File.expand_path("..", __FILE__)
                unless error.backtrace.frozen?
                  error.backtrace.reject! { |line| line.start_with?(lib) }
                  error.backtrace_locations.reject! { |line| line.path.start_with?(lib) }
                end
              end
```

lib/spring/application.rb
```diff
@@ -319,7 +319,10 @@ def shush_backtraces
             ensure
```
amomchilov, line 319: Found this confusing. Is there any reason to use `ensure` + `$!`, rather than just `rescue => e` + `raise e`?

I thought perhaps because a re-raise would introduce a new frame into the backtrace, but it doesn't

<details><summary>demo</summary>

```rb
#!/opt/rubies/3.4.2/bin/ruby

begin
 raise
rescue => e
  raise e
end

# Prints:
# demo.rb:4:in '<main>': unhandled exception

# As you see, line 5 and 6 aren't in the backtrace.
```

</details>

lib/spring/application.rb
```diff
@@ -319,7 +319,10 @@ def shush_backtraces
...
               if $!
                 lib = File.expand_path("..", __FILE__)
-                $!.backtrace.reject! { |line| line.start_with?(lib) } unless $!.backtrace.frozen?
+                unless $!.backtrace.frozen?
+                  $!.backtrace.reject! { |line| line.start_with?(lib) }
+                  $!.backtrace_locations.reject! { |line| line.path.start_with?(lib) }
```
byroot, line 324:
```suggestion
                  $!.backtrace_locations.reject! { |line| line.path&.start_with?(lib) }
```

Pretty sure `path` can be `nil` if the method is implemented in C.

byroot: Also, while it's very convoluted, `.backtrace` could be mutable, but `.backtrace_location` frozen.

Issue #676, opened 2022-05-20
Different exit code for SIGKILL'ed processes with Spring


noizwaves:
With Spring enabled, a Ruby command interrupted by a signal exits with a **different exit code** compared to that same command run with Spring disabled.

This issue was detected in our containerized development environments, which run inside of Docker. Because Docker enforces memory constraints by OOM'ing and SIGKILL'ing offending processes. The differing behavior can be observed in an application (Ruby 2.7.5, Rails 7.0.2.4, Spring 4.0.0) when forcing an OOM error:

**Spring enabled:**
```bash
$ bin/rails runner "arr = []; while true do; arr.push(1); end"
Running via Spring preloader in process 242
$ echo $?
0
```

**Spring disabled:**
```bash
$ DISABLE_SPRING=1 bin/rails runner "arr = []; while true do; arr.push(1); end"
Killed
$ echo $?
137
```

It looks like the `status.exitstatus` returned by `Process.wait2` is `nil` for signaled processes, and the conversion to the [corresponding unix/POSIX exit code](https://unix.stackexchange.com/a/99134) for that signal is not done implicitly. `exitstatus` is returned directly by terminated processes [here](https://github.com/rails/spring/blob/v4.0.0/lib/spring/application.rb#L345).

Could this conversion of signal to exit code be added?

PR #741, opened 2025-06-20
Pass signaled exit code properly to the client

zygzagZ:
Pass signaled exit code properly to the client

Process::Status#existstatus is nil when child did not exit cleanly. When ruby process crashes, running it with spring masked exit code and returned 0.
This commit allows Spring::Server thread to properly pass application exit code to the client, even when signaled or stopped.

Fixes https://github.com/rails/spring/issues/676.


rafaelfranca:
Closing in favor of #744

PR #751, opened 2026-04-28
Don't crash on client disconnect (Server and Application)

### LINKED ISSUE #724: Client timeout causes server crash
I noticed recently that when a Spring client times out, that it also crashes the server. This appears to be because the server tries to write to the client and gets a `Broken pipe` error. In this case I would expect the server to remain up.

Client:
```
/app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/lib/spring/client/run.rb:124:in `verify_server_version': Error connecting to Spring server (RuntimeError)
        from /app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/lib/spring/client/run.rb:60:in `run'
        from /app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/lib/spring/client/run.rb:40:in `warm_run'
        from /app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/lib/spring/client/run.rb:33:in `call'
        from /app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/lib/spring/client/command.rb:7:in `call'
        from /app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/lib/spring/client/rails.rb:22:in `call'
        from /app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/lib/spring/client/command.rb:7:in `call'
        from /app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/lib/spring/client.rb:30:in `run'
        from /app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/bin/spring:49:in `<top (required)>'
```

Server:
```
[2024-08-02 21:13:08 +0000] [1] [application_manager:development] got worker pid 287
[2024-08-02 21:13:08 +0000] [198] [client] got pid: 287
[2024-08-02 21:13:08 +0000] [1] [server] accepted client
[2024-08-02 21:13:08 +0000] [1] [server] shutting down
[2024-08-02 21:13:08 +0000] [1] [application_manager:development] stopping
[2024-08-02 21:13:08 +0000] [207] [application:development] running -> terminating
[2024-08-02 21:13:08 +0000] [207] [application:development] terminating -> exiting
[2024-08-02 21:13:09 +0000] [207] [application:development] 287 exited with 0
[2024-08-02 21:13:09 +0000] [198] [client] got exit status 0
/app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/lib/spring/server.rb:57:in `write': Broken pipe (Errno::EPIPE)
        from /app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/lib/spring/server.rb:57:in `puts'
        from /app/vendor/bundle/ruby/3.3.0/gems/spring-4.2.1/lib/spring/server.rb:57:in `serve'
```

Korri:
Fixes #724. Same bug as the closed #725, with the additional crash sites that PR missed.

When a client disconnects mid-handshake — typically [`Spring.connect_timeout`](https://github.com/rails/spring/blob/3347006f576db6e6e72e9a1549a3e5b3311307aa/lib/spring/configuration.rb) (default 5s) firing on a slow preloader boot, reproduced reliably on Shopify's monorepo — every unrescued write to the client raises `Errno::EPIPE` and kills the process holding it. There are 5 such writes across 2 processes:

- 1 in [`Spring::Server#serve`](https://github.com/rails/spring/blob/3347006f576db6e6e72e9a1549a3e5b3311307aa/lib/spring/server.rb#L57) — **worst case**: the accept loop dies, the unix socket lingers, every subsequent client times out in `verify_server_version` until `spring stop`.
- 4 in [`Spring::Application#serve`](https://github.com/rails/spring/blob/3347006f576db6e6e72e9a1549a3e5b3311307aa/lib/spring/application.rb#L181-L260): main-flow `client.puts(0)` + three writes nested *inside* the existing [`rescue Exception`](https://github.com/rails/spring/blob/3347006f576db6e6e72e9a1549a3e5b3311307aa/lib/spring/application.rb#L248) handler (a sibling rescue at the same level can't catch them).

Fixes adds a few `rescue Errno::EPIPE` to gracefully log or recover if the client is gone.

## Open questions

- The `client.puts(1) if pid` rescue covers the post-fork exit-status write; reproducing it in a unit test would require lots of stubbing, I left it out of tests.
- I considered also rescuing `Errno::ECONNRESET`, but kept the surface to what #724 actually reports. Easy to broaden.
- Pre-existing crash at [`application.rb:164`](https://github.com/rails/spring/blob/3347006f576db6e6e72e9a1549a3e5b3311307aa/lib/spring/application.rb#L164) (`manager.recv_io` raising `SocketError` on manager close) is a different path and out of scope here — worth a follow-up.

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -185,7 +185,7 @@ module Spring
           client.puts(0) # preload success
         rescue Exception
           log "preload failed"
-          client.puts(1) # preload failure
+          ignore_client_disconnect { client.puts(1) } # preload failure
           raise
         end
       end
@@ -246,15 +246,19 @@ module Spring
 
       wait pid, streams, client
     rescue Exception => e
-      log "exception: #{e}"
+      if e.is_a?(Errno::EPIPE)
+        log "client disconnected (#{e.message}), ignoring command"
+      else
+        log "exception: #{e}"
+      end
       manager.puts unless pid
 
       if streams && !e.is_a?(SystemExit)
-        print_exception(stderr, e)
-        streams.each(&:close)
+        ignore_client_disconnect { print_exception(stderr, e) }
+        streams.each { |stream| ignore_client_disconnect { stream.close } }
       end
 
-      client.puts(1) if pid
+      ignore_client_disconnect { client.puts(1) if pid }
       client.close
     ensure
       # Redirect STDOUT and STDERR to prevent from keeping the original FDs
@@ -410,6 +414,15 @@ module Spring
     end
 
     private
+
+    # Tolerate Errno::EPIPE on writes to the client socket. Once the client
+    # disconnects, every subsequent write to it raises — that's expected
+    # during `serve`'s status reporting; we'd be talking to a process
+    # that's gone.
+    def ignore_client_disconnect
+      yield
+    rescue Errno::EPIPE
+    end
 
     def active_record_configured?
       defined?(ActiveRecord::Base) && ActiveRecord::Base.configurations.any?
--- a/test/unit/application_test.rb
+++ b/test/unit/application_test.rb
@@ -1,0 +1,110 @@
+require_relative "../helper"
+require "spring/env"
+require "spring/application"
+require "timeout"
+
+# Regression tests for https://github.com/rails/spring/issues/724:
+# disconnected client raises Errno::EPIPE in #serve, which used to crash
+# the application process. Each test exercises one of the EPIPE call
+# sites in `Spring::Application#serve` with a `UNIXSocket.pair` whose
+# client side has been closed.
+class ApplicationTest < ActiveSupport::TestCase
+  def setup
+    @log_file = File.open(File::NULL, "a")
+    @spring_env = Spring::Env.new(log_file: @log_file)
+    @manager_read, @manager_write = UNIXSocket.pair
+    @app = Spring::Application.new(@manager_write, {}, @spring_env)
+    @server_sock = build_disconnected_application_client
+  end
+
+  def teardown
+    @server_sock&.close
+    @manager_read&.close
+    @manager_write&.close
+    @log_file&.close
+  end
+
+  test "#serve does not raise when the client disconnects before the cached-preload-success write" do
+    @app.define_singleton_method(:preloaded?) { true }
+
+    with_saved_stdio { @app.serve(@server_sock) }
+
+    assert_manager_handshake_complete
+  end
+
+  test "#serve does not raise when the client disconnects before the fresh-preload-success write" do
+    @app.define_singleton_method(:preload) {}
+
+    with_saved_stdio { @app.serve(@server_sock) }
+
+    assert_manager_handshake_complete
+  end
+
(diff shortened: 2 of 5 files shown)


lib/spring/application.rb
```diff
@@ -183,7 +183,11 @@ def serve(client)
           client.puts(0) # preload success
         rescue Exception
           log "preload failed"
-          client.puts(1) # preload failure
+          begin
+            client.puts(1) # preload failure
+          rescue Errno::EPIPE
```
rafaelfranca: This is too duplicated. Can we remove this duplication?

test/unit/application_test.rb
```diff
@@ -0,0 +1,165 @@
...
+    manager_read, manager_write = UNIXSocket.pair
+
+    app = Spring::Application.allocate
+    app.instance_variable_set(:@spring_env, spring_env)
+    app.instance_variable_set(:@manager,    manager_write)
+    app.instance_variable_set(:@clients,    {})
+    app.instance_variable_set(:@preloaded,  preloaded)
+    app.instance_variable_set(:@state,      :running)
+    app.instance_variable_set(:@mutex,      Mutex.new)
+    app.instance_variable_set(:@waiting,    {})
```
rafaelfranca: This makes no sense. Our tests should not be breaking encapsulation. Can we write tests that makes sense please?

test/unit/server_test.rb
```diff
@@ -0,0 +1,38 @@
...
+  def build_server
+    fake_env = Object.new
+    fake_env.define_singleton_method(:version)  { "test-version" }
+    fake_env.define_singleton_method(:log)      { |_message| }
+    fake_env.define_singleton_method(:log_file) { File::NULL }
+
+    server = Spring::Server.allocate
+    server.instance_variable_set(:@env, fake_env)
+    server.define_singleton_method(:redirect_output) {}
+    server
```
rafaelfranca: Same here, this is breaking encapsulation

lib/spring/application.rb
Korri: Pulled the four `begin / rescue Errno::EPIPE / end` blocks into a single private `ignore_client_disconnect` helper.

test/unit/application_test.rb
Korri: Switched to the real constructor:

```ruby
@spring_env = Spring::Env.new(log_file: @log_file)
@manager_read, @manager_write = UNIXSocket.pair
@app = Spring::Application.new(@manager_write, {}, @spring_env)
```

Plus some more targeted stubbing.

test/unit/server_test.rb
Korri: Same fix as above 👍🏻

test/unit/application_test.rb
Korri: That makes for quite a verbose test setup/structure, open to suggestions on better approaches, it's tricky given how close to the metal the changes are.

PR #752, opened 2026-04-29
Eagerly preload framework base classes

Korri:
`Spring::Application#serve` calls `Rails.application.reloaders.any?(&:updated?)` before forking. ActionView's [`CacheExpiry::ViewReloader#updated?`](https://github.com/rails/rails/blob/main/actionview/lib/action_view/cache_expiry.rb) lazily builds a `FileUpdateChecker` from whatever view paths are registered at that moment (lazy-init contract from rails/rails#51308).

Engines register view paths via `prepend_view_path` inside `on_load(:action_mailer)` / `:action_controller_base` blocks. In a fresh preload nothing has referenced those framework classes yet, so `dirs_to_watch` is empty when the probe runs — pinning the watcher empty and triggering a full `FileUpdateChecker` rebuild on every `prepend_view_path` in each fork.

Preloading the framework classes (not `Application*`, per Rafael's suggestion, framework code is never reloaded, app namespace is) targets the `on_load` hooks directly and avoids coupling Spring's preload to user-reloadable code.

While [rails/rails#57269](https://github.com/rails/rails/pull/57269) closes the bug part of this from the Rails side, this still makes sense on it's own.

### DIFF
--- a/lib/spring/application.rb
+++ b/lib/spring/application.rb
@@ -117,6 +117,7 @@ module Spring
       require Spring.application_root_path.join("config", "environment")
 
       invoke_after_environment_load_callbacks
+      preload_framework_base_classes
 
       disconnect_database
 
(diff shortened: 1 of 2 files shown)


lib/spring/application.rb
```diff
@@ -142,6 +144,29 @@ def preload
...
+    # had a chance to register their view paths. The empty watcher then
+    # gets pinned in a non-lazy state and triggers a full FileUpdateChecker
+    # rebuild on every prepend_view_path inside each fork.
+    #
+    # See rails/rails#51308 for the lazy-init contract this aligns with.
+    APPLICATION_BASE_CLASSES = %w[
+      ApplicationController
+      ApplicationMailer
+      ApplicationJob
+      ApplicationRecord
```
rafaelfranca: I think it would be better to load the framework code, not the application code. The application code can change and is reloaded, but if you keep a reference in the server that might break thing. The framework code can never change.

PR #756, opened 2026-06-05
Make client smarter about Spring server state

Korri:
This covers two ways a client can find an existing Spring socket but still fail the initial version handshake:

   - another client booted the server first, and the server is still busy preloading the app before it can accept this client
       - ➡️ now we retry with `timeout: Spring.boot_timeout`.
   - the previous server died after leaving its socket/pidfile on disk
       - ➡️ now we check for the existence of the process, and start a new one if it's not there. 

Before this, both could surface as “Error connecting to Spring server” even though the right behavior is recoverable: wait a little longer for the live booting server, or restart when the old server is gone.

The retry is still bounded. If the `pidfile` lock says a server is alive, a client that did not boot it gets one `boot_timeout`-sized wait for the greeting. If this client already booted the server, or the extra wait still does not produce a greeting, we keep failing/restarting instead of looping forever.

I added acceptance coverage with real Spring processes for both cases, including a probe that proves the second client actually took the boot-timeout path.

### DIFF
--- a/lib/spring/client/run.rb
+++ b/lib/spring/client/run.rb
@@ -6,6 +6,7 @@ module Spring
   module Client
     class Run < Command
       FORWARDED_SIGNALS = %w(INT QUIT USR1 USR2 INFO WINCH) & Signal.list.keys
+      ServerReadTimeout = Class.new(StandardError)
 
       attr_reader :server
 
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Tests are failing with a timeout. Can you take a look?

Korri:
> Tests are failing with a timeout. Can you take a look?

@rafaelfranca should be good now, the new tests had some leakage.

